PR #1, opened 2021-05-05
8264311: Heap object statistics

rkennke:
For Lilliput evaluation, it would be useful to be able to get some statistics about heap objects, e.g. how many objects there typically are, what is their (average) size, how big is the live data set, how many objecs have an identity hash code and how many objects are locked. Some of that information may be useful for general purpose too, e.g. the avg live data set and object size/count information might be quite useful to have.

Heap object statistics can be gathered and printed by invoking java with -XX:+UnlockDiagnosticVMOptions -XX:+HeapObjectStats -Xlog:heap+stats. It will impact performance of the workload. The sampling interval can be specificed by -XX:HeapObjectStatsSamplingInterval=X (in ms) and defaults to 500 (ms).

Testing:
 - [x] Some manual tests, verifying the plausability of the output by hand
 - [x] tier1
 - [x] tier2


src/hotspot/share/services/heapObjectStatistics.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+class HeapObjectStatisticsTask : public PeriodicTask {
+public:
+  HeapObjectStatisticsTask();
+  void task();
+};
+
+class HeapObjectStatistics : public CHeapObj<mtGC> {
+private:
+  static const int HISTOGRAM_SIZE = 16;
```
shipilev, line 43: This fields looks unused.

shipilev (approved):

shipilev (approved):

shipilev:
I think my review does not count here, because I am in "Commiter" role here. You can relax this by doing e.g. https://github.com/openjdk/jcstress/blob/master/.jcheck/conf#L39-L41

rkennke:
> I think my review does not count here, because I am in "Commiter" role here. You can relax this by doing e.g. https://github.com/openjdk/jcstress/blob/master/.jcheck/conf#L39-L41

Right. I fixed it in master branch, and rebased this PR on the master. Let's do a last attempt at getting this 'properly reviewed' and failing that I'll bypass the bot.

shipilev (approved):

PR #3, opened 2021-05-10
Trim identity-hashcode to 25bit

rkennke:
I'd like to trim the identity hashcode to 25 bits to make room in the upper 32bits for the compressed class-pointer.

This could in theory regress performance on workloads that make heavy use of i-hashes, but I could not show this in practice (by specjvm workloads like compiler which uses lots of ihashes).

Next step would be to move the Klass* into the upper 32bit part. (If you look closely, this would be the exact header layout of 32bit JVMs.) And then get rid of the dedicated Klass* field. :-D Eventually I'd make ihash-bits and klass* bits configurable to trade one for the other.

Testing:
 - [x] some manual benchmarks
 - [x] tier1
 - [ ] tier2

---------
### Progress

### Reviewers
 * [Aleksey Shipilev](https://openjdk.java.net/census#shade) (@shipilev - Committer) ⚠️ Review applies to c3790d0cd45a65491a522a4645f5e09814dbc779

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/3/head:pull/3` \
`$ git checkout pull/3`

`$ git checkout pull/3` \
`$ git pull https://git.openjdk.java.net/lilliput pull/3/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 3`

`$ git pr show -t 3`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/3.diff">https://git.openjdk.java.net/lilliput/pull/3.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/oops/markWord.hpp
+++ b/src/hotspot/share/oops/markWord.hpp
@@ -41,8 +41,8 @@
 //
 //  64 bits:
 //  --------
-//  unused:25 hash:31 -->| unused_gap:1   age:4    biased_lock:1 lock:2 (normal object)
-//  JavaThread*:54 epoch:2 unused_gap:1   age:4    biased_lock:1 lock:2 (biased object)
+//  unused:32 hash:25 -->|  age:4    biased_lock:1 lock:2 (normal object)
+//  JavaThread*:54 epoch:2  age:4    biased_lock:1 lock:2 (biased object)
 //
 //  - hash contains the identity hash value: largest value is
 //    31 bits, see os::random().  Also, 64-bit vm's require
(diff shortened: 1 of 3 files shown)


src/hotspot/share/oops/markWord.hpp
```diff
@@ -41,8 +41,8 @@
 //
 //  64 bits:
 //  --------
-//  unused:25 hash:31 -->| unused_gap:1   age:4    biased_lock:1 lock:2 (normal object)
-//  JavaThread*:54 epoch:2 unused_gap:1   age:4    biased_lock:1 lock:2 (biased object)
+//  unused:32 hash:25 -->|  age:4    biased_lock:1 lock:2 (normal object)
+//  JavaThread*:54 epoch:2  age:4    biased_lock:1 lock:2 (biased object)
```
shipilev, line 45: This does not compute. Where did the `unused_gap` go in the second case? Is `epoch` now larger? Or does `JavaThread*` now takes 55 bits?

rkennke: Epoch and JavaThread in the biased case are already overlaid with hashcode. Hashcode and biased state doesn't happen at the same time. 32+25+4+1+2 == 64.

shipilev: Right, but I am asking about the _second case_, though, where: 54+2+4+1+2 == 63. Where does the `unused_gap` bit go?

PR #6, opened 2021-05-18
Unconditionally preserve all marks during GC

rkennke:
When storing the Klass* in the object header, we need to preserve all marks during GC, because otherwise we'd be loosing the Klass*.

This means that the PreservedMarks structure is repurposed as a full (reverse) forwarding table. Reverse because instead of mapping object->forwardee, and leaving the mark alone, it maps object->mark and stores the forwardee in the header, which might be a better idea because it means we can have a most compact table.

Shenandoah doesn't have this problem, except during full-GC, and ZGC has its own forwarding table.

This change will likely affect performance, but I haven't checked by how much, yet. It's probably more useful to see this in the context of actually reduced header.
Testing:
 - [x] tier1
 - [x] tier2

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/6/head:pull/6` \
`$ git checkout pull/6`

`$ git checkout pull/6` \
`$ git pull https://git.openjdk.java.net/lilliput pull/6/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 6`

`$ git pr show -t 6`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/6.diff">https://git.openjdk.java.net/lilliput/pull/6.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/g1/g1FullGCCompactionPoint.cpp
+++ b/src/hotspot/share/gc/g1/g1FullGCCompactionPoint.cpp
@@ -113,13 +113,6 @@ void G1FullGCCompactionPoint::forward(oop object, size_t size) {
       // even if the mark-word is used. This is no problem since
       // forwardee() will return NULL in the compaction phase as well.
       object->init_mark();
-    } else {
-      // Make sure object has the correct mark-word set or that it will be
-      // fixed when restoring the preserved marks.
-      assert(object->mark() == markWord::prototype_for_klass(object->klass()) || // Correct mark
-             (UseBiasedLocking && object->has_bias_pattern()), // Will be restored by BiasedLocking
-             "should have correct prototype obj: " PTR_FORMAT " mark: " PTR_FORMAT " prototype: " PTR_FORMAT,
-             p2i(object), object->mark().value(), markWord::prototype_for_klass(object->klass()).value());
     }
     assert(object->forwardee() == NULL, "should be forwarded to NULL");
   }
--- a/src/hotspot/share/gc/g1/g1FullGCCompactionPoint.cpp
+++ b/src/hotspot/share/gc/g1/g1FullGCCompactionPoint.cpp
@@ -117,7 +117,6 @@ void G1FullGCCompactionPoint::forward(oop object, size_t size) {
       // Make sure object has the correct mark-word set or that it will be
       // fixed when restoring the preserved marks.
       assert(object->mark() == markWord::prototype_for_klass(object->klass()) || // Correct mark
-             object->mark_must_be_preserved() || // Will be restored by PreservedMarksSet
              (UseBiasedLocking && object->has_bias_pattern()), // Will be restored by BiasedLocking
              "should have correct prototype obj: " PTR_FORMAT " mark: " PTR_FORMAT " prototype: " PTR_FORMAT,
              p2i(object), object->mark().value(), markWord::prototype_for_klass(object->klass()).value());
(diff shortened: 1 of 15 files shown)


src/hotspot/share/gc/g1/g1FullGCCompactionPoint.cpp
```diff
@@ -117,7 +117,6 @@ void G1FullGCCompactionPoint::forward(oop object, size_t size) {
       // Make sure object has the correct mark-word set or that it will be
       // fixed when restoring the preserved marks.
       assert(object->mark() == markWord::prototype_for_klass(object->klass()) || // Correct mark
-             object->mark_must_be_preserved() || // Will be restored by PreservedMarksSet
```
shipilev, line 120: It would seem this assert checks that mark is correct or it would be preserved. Since all marks are now preserved, this assert is basically `true` at all times? If you just drop `object->mark_must_be_preserved()`, it should be failing with forwarded objects, no?

src/hotspot/share/gc/g1/g1FullGCOopClosures.inline.hpp
```diff
@@ -80,7 +80,6 @@ template <class T> inline void G1AdjustClosure::adjust_pointer(T* p) {
   if (forwardee == NULL) {
     // Not forwarded, return current reference.
     assert(obj->mark() == markWord::prototype_for_klass(obj->klass()) || // Correct mark
-           obj->mark_must_be_preserved() || // Will be restored by PreservedMarksSet
```
shipilev, line 83: Same as above.

shipilev (changes requested):
I have questions about the asserts.

src/hotspot/share/gc/g1/g1FullGCCompactionPoint.cpp
rkennke: This block concerns about non-forwarded objects because forwarded objects are handled above.

src/hotspot/share/gc/g1/g1FullGCOopClosures.inline.hpp
rkennke: Yeah, same as above ;-)

src/hotspot/share/gc/g1/g1FullGCCompactionPoint.cpp
shipilev: OK, but I am still confused what `mark_must_be_preserved` case does in the original assert. Faithfully, you would need to replace that case with `true`, and then the whole assert should fold to always passing? Maybe it handles identityHashCode-bearing or displaced-header objects?

rkennke: Right. Originally, only relatively few headers need to be preserved. Specifically, all headers that are involved in locking or which carry a hash-code. The assert checks that we either have a header that gets preserved, or a default/prototype header. Now that all headers are preserved, the asserts fold-up to be always true and can go away. I'll fix that.

shipilev (approved):
Looks fine to me as experimental code.

rkennke:
> Looks fine to me as experimental code.

Thanks.
However, I noticed that we (will) have another problem during GC: For sliding compaction, while objects are forwarded, we will also require the Klass* *at the same time* e.g. for heap traversal. Which means, we cannot simply override the Klass* in the header, not even temporarily. Solutions to this could be avoiding klass-based traversal (e.g. using bitmaps, but serial/parallel GC don't use a bitmap), or an external forwarding table. If we go for the forwarding table, which is what I'm currently leaning to, we don't need the preserved-marks anymore. I need to think this through a little bit.

rkennke:
I am closing this because it is no longer needed.

PR #8, opened 2021-06-10
Implement sliding forwarding scheme that preserves upper header bits

rkennke:
The current way of storing forwarding information - by storing the forwardee address into the header - poses a problem for sliding collectors: it overrides the upper bits that keep the (compressed) Klass*. Sliding collectors have no way to recover the Klass*, and therefore need a different forwarding scheme that preserves the upper bits. I propose a scheme that compresses the forwarding pointer and by taking advantage of the fact that each region only ever forwards to at most two other regions (when dividing the heap into equal-sized logical regions), it can address all of the heap, regardless of its size. This obviouly works well with regionalized collectors, but it also works with contiguous-heap collectors like serial or parallel GC. The latter would divide the heap into logical regions sized by 4G (or single region if heap is smaller than that). Notice that evacuating and scavenging collectors don't have this problem: they can safely stomp over the from-space copy of objects, the Klass* information is preserved in the to-space copy.

G1 is special: the fallback from parallel to serial full GC means that we can have maximum of N target regions with N == ParallelGCThreads. For this reason, I use an encoding with 5 bits for target region index, that is enough to address 32 target regions from each region, assuming a maximum heap region size of 32M. In the future we will be able to steal bits from the Klass* in the upper half of the header, allowing us to address even more regions. Alternatively, we can change the full-GC fallback to re-forward all objects serially, instead of trying to re-use the already-forwarded state from the parallel full GC.

Serial GC and Shenandoah GC use only 1 bit to address target regions, because they really can only ever have 2 target regions out of a single region.

Testing:
- [x] manual testing with +passive and -degen-gc
- [x] hotspot_gc_shenandoah
- [x] tier1
- [x] tier2
- [x] tier1 (x86_32)

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/8/head:pull/8` \
`$ git checkout pull/8`

`$ git checkout pull/8` \
`$ git pull https://git.openjdk.java.net/lilliput pull/8/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 8`

`$ git pr show -t 8`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/8.diff">https://git.openjdk.java.net/lilliput/pull/8.diff</a>

</details>


src/hotspot/share/gc/g1/g1FullGCOopClosures.inline.hpp
```diff
@@ -87,8 +88,8 @@ template <class T> inline void G1AdjustClosure::adjust_pointer(T* p) {
   }
 
   // Forwarded, just update.
-  oop forwardee = obj->forwardee();
-  assert(G1CollectedHeap::heap()->is_in_reserved(forwardee), "should be in object space");
+  oop forwardee = _forwarding->forwardee(obj);
+  assert(G1CollectedHeap::heap()->is_in_reserved(forwardee), "should be in object space: " PTR_FORMAT "(" PTR_FORMAT ", " PTR_FORMAT ") " INTPTR_FORMAT, p2i(forwardee), p2i(G1CollectedHeap::heap()->reserved().start()), p2i(G1CollectedHeap::heap()->reserved().end()), obj->mark().value());
```
shipilev, line 92: Style: break the line here.

src/hotspot/share/gc/g1/g1FullCollector.cpp
```diff
@@ -300,13 +301,15 @@ void G1FullCollector::phase1_mark_live_objects() {
...
   GCTraceTime(Info, gc, phases) info("Phase 2: Prepare for compaction", scope()->timer());
+  _heap->forwarding()->clear();
   G1FullGCPrepareTask task(this);
   run_task(&task);
 
   // To avoid OOM when there is memory left.
-  if (!task.has_freed_regions()) {
-    task.prepare_serial_compaction();
-  }
+  // NOTE: Disabled for now because it violates sliding-forwarding assumption.
```
shipilev, line 309: I think it should be "TODO", so source analyzers can highlight this.

src/hotspot/share/gc/shared/preservedMarks.hpp
```diff
@@ -65,6 +66,8 @@ class PreservedMarks {
   // to their forwarding location stored in the mark.
   void adjust_during_full_gc();
```
shipilev, line 67: Does this method have any uses? Should those be updated to use `forwarding`? If not, it should be removed for safety?

src/hotspot/share/gc/shared/slidingForwarding.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * SlidingForwarding is a method to store forwarding information in a compressed form into the object header,
+ * that has been specifically designed for sliding compaction GCs.
+ * It avoids overriding the compressed class pointer in the upper bits of the header, which would otherwise
+ * be lost. SlidingForwarding requires only small side tables and guarantees constant-time access and modification.
+ *
+ * The idea is to use a pointer compression scheme very similar to the one that is used for compressed oops.
+ * We divide the heap into number of logical regions. Each region spans maximum of 2^NUM_BITS words.
+ * We take advantage of the fact that sliding compaction can forward objects from ore region to a maximum of
```
shipilev, line 40:
```suggestion
 * We take advantage of the fact that sliding compaction can forward objects from one region to a maximum of
```

src/hotspot/share/gc/shared/slidingForwarding.inline.hpp
```diff
@@ -0,0 +1,110 @@
...
+}
+
+
+uintptr_t SlidingForwarding::encode_forwarding(HeapWord* original, HeapWord* target) {
+  size_t orig_idx = region_index_containing(original);
+  size_t base_table_idx = orig_idx * 2;
+  size_t target_idx = region_index_containing(target);
+  HeapWord* encode_base;
+  uintptr_t region_idx;
+  for (region_idx = 0; region_idx < (ONE << NUM_REGION_BITS); region_idx++) {
```
shipilev, line 50: I think `ONE << NUM_REGION_BITS` really deserves a separate constant.

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
```diff
@@ -370,7 +374,7 @@ class ShenandoahPrepareForCompactionTask : public AbstractGangTask {
   ShenandoahHeapRegionSet** const _worker_slices;
 
 public:
-  ShenandoahPrepareForCompactionTask(PreservedMarksSet *preserved_marks, ShenandoahHeapRegionSet **worker_slices) :
+  ShenandoahPrepareForCompactionTask(PreservedMarksSet* preserved_marks, ShenandoahHeapRegionSet **worker_slices) :
```
shipilev, line 377: Irrelevant change.

test/hotspot/jtreg/gc/stress/TestMultiThreadStressRSet.java
```diff
@@ -49,7 +49,7 @@
  *
  * @run main/othervm -Xbootclasspath/a:. -XX:+UnlockDiagnosticVMOptions -XX:+WhiteBoxAPI
  *   -XX:+UseG1GC -XX:G1SummarizeRSetStatsPeriod=100 -Xlog:gc
- *   -Xmx1G -XX:G1HeapRegionSize=8m -XX:MaxGCPauseMillis=1000 gc.stress.TestMultiThreadStressRSet 60 16
+ *   -Xmx1100M -XX:G1HeapRegionSize=8m -XX:MaxGCPauseMillis=1000 gc.stress.TestMultiThreadStressRSet 60 16
```
shipilev, line 52: Why this change?

shipilev (approved):
Okay, I admit this is cute! I have a few cosmetic comments, but otherwise, it looks good for the experimental code. I would probably have a round of code tideups later...

src/hotspot/share/gc/g1/g1FullGCOopClosures.inline.hpp
rkennke: I'd rather remove the new debug output.

src/hotspot/share/gc/shared/preservedMarks.hpp
rkennke: Actually not, except in a unit test. This test should be written to use a forwarding instead, but this probably requires a mockable/abstract forwarding that can be plugged in for the purpose of the test. I added TODOs and postpone this to later.

test/hotspot/jtreg/gc/stress/TestMultiThreadStressRSet.java
rkennke: This is related to disabling the last-last-ditch serial full GC. Apparently, in the old configuration, the serial-full-GC would free enough remaining regions to make this test work. Without the change, it runs into OOM.

shipilev (approved):

PR #12, opened 2021-07-13
Read class from object header

rkennke:
This changes the Hotspot runtime to load the Klass* from the header instead of the dedicated Klass* word. The dedicated word is only still used for verification and for access by generated code (the former will eventually go away, the latter will be implemented separately).

Currently, this means we need to coordinate with the ObjectSynchronizer: when encountering a header that is a stack lock or a monitor, the header is displaced. Worse, if it is a stack-locked that is owned by a thread other than the calling thread, we must first inflate the lock to a full monitor. This is particularily bad for GCs. Luckily, most paths only do this at a safepoint, where we can actually safely access foreign stack locks and don't need to worry about inflation. Notably exception is concurrent marking by G1GC, which can cause inflation of locks, but it doesn't hurt very much.

It's really bad for Shenandoah and ZGC, though: when relocating objects, GC needs to know the object size of the from-space copy. However, this can cause inflation, and inflation creates new WeakHandle in the resulting monitor, and that would be initialized with a from-space copy, which is a no-go during evacuation/relocation.

That said, I have been told that work is under way to get rid of displaced headers altogether, which would neatly solve all those problems. I have no desire to make complicated workarounds for Shenandoah GC and ZGC. I disabled both in my own builds for now, and will implement them as soon as the monitor changes arrive.

In a couple of places in GC we need to access the header carefully: when concurrently forwarding (by parallel GC threads), we need to ensure we access the Klass* from an unforwarded header, and must also ensure to avoid re-loading the Klass* once we have the good header (that is why so many asserts have been removed - they would potentially re-load the Klass* from a header that may now be forwarded).

Testing:
 - [x] tier1
 - [x] tier2
 - [x] hotspot_gc
(all without Shenandoah and ZGC, see above)

---------
### Progress

### Reviewers
 * [Aleksey Shipilev](https://openjdk.java.net/census#shade) (@shipilev - Committer) ⚠️ Review applies to 1f9e5aef7558b950aef1a53a7a9aef5f549edd3b

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/12/head:pull/12` \
`$ git checkout pull/12`

`$ git checkout pull/12` \
`$ git pull https://git.openjdk.java.net/lilliput pull/12/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 12`

`$ git pr show -t 12`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/12.diff">https://git.openjdk.java.net/lilliput/pull/12.diff</a>

</details>


src/hotspot/share/oops/oop.inline.hpp
```diff
@@ -80,39 +82,63 @@ markWord oopDesc::cas_set_mark(markWord new_mark, markWord old_mark, atomic_memo
...
 Klass* oopDesc::klass_or_null() const {
-  if (UseCompressedClassPointers) {
-    return CompressedKlassPointers::decode(_metadata._compressed_klass);
-  } else {
-    return _metadata._klass;
+#ifdef _LP64
+  assert(UseCompressedClassPointers, "only with compressed class pointers");
+  markWord header = mark();
+  if (!header.is_neutral()) {
+    header =ObjectSynchronizer::stable_mark(cast_to_oop(this));
```
shipilev, line 117:
```suggestion
    header = ObjectSynchronizer::stable_mark(cast_to_oop(this));
```

src/hotspot/share/oops/markWord.inline.hpp
```diff
@@ -0,0 +1,66 @@
...
+markWord markWord::set_narrow_klass(const narrowKlass nklass) const {
+  return markWord((value() & ~klass_mask_in_place) | ((uintptr_t) nklass << klass_shift));
+}
+
+Klass* markWord::safe_klass() const {
+  assert(SafepointSynchronize::is_at_safepoint(), "only call at safepoint");
+  markWord mrk = *this;
+  if (mrk.has_displaced_mark_helper()) {
+    mrk = mrk.displaced_mark_helper();
+  }
```
shipilev, line 54:
```suggestion
  markWord m = *this;
  if (m.has_displaced_mark_helper()) {
    m = m.displaced_mark_helper();
  }
```

src/hotspot/share/oops/oop.hpp
```diff
@@ -77,6 +77,7 @@ class oopDesc {
   inline Klass* klass_or_null() const;
   inline Klass* klass_or_null_acquire() const;
 
+  narrowKlass narrow_klass() const { return _metadata._compressed_klass; }
```
shipilev, line 80: Maybe we should rename this to e.g. `narrow_klass_legacy` to avoid accidentally using it? AFAIU, this is for legacy code paths?

shipilev (approved):
All right, it looks fine for the experimental code. A few questions/comments:

rkennke:
> The following ? ?Compact Java Monitors? -- might provide some relief : https://arxiv.org/pdf/2102.04188.pdf.
> 
> As described, it handles just the identity hashCode value, but it?s trivial to extend the idea to displacing the whole lilliput header word.

Thanks, Dave! I will study it!
My current thinking wrt identityHashcode is to not store it in the header at all, but instead append it to the object on-demand. That is, when identityHashCode() is called, it would get re-computed as long as the object does not move, and as soon as it moves, an extra field would get appended to the object (or rather, often it fits in alignment gap at the end), and the hashcode is installed there. This is implemented in prototype here: https://github.com/rkennke/lilliput/tree/compact-hashcode
But, as far as I can see, it doesn't help with the problem that concurrent GCs have with the thread-locks.

src/hotspot/share/gc/shared/preservedMarks.inline.hpp
```diff
@@ -48,7 +48,19 @@ inline void PreservedMarks::push_if_necessary(oop obj, markWord m) {
...
 inline void PreservedMarks::init_forwarded_mark(oop obj) {
-  obj->init_mark();
+  assert(obj->is_forwarded(), "only forwarded here");
+#ifdef _LP64
+  oop forwardee = obj->forwardee();
+  markWord header = forwardee->mark();
+  if (header.has_displaced_mark_helper()) {
+    header = header.displaced_mark_helper();
+  }
+  narrowKlass nklass = header.narrow_klass();
```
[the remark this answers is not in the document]
rkennke: Yes, the whole Lilliput assumes UseCompressedClassPointers==true. We enforce that in arguments.cpp. We might even go so far to remove the flag and code paths altogether, but not now.

rkennke:
> I looked through compact-hashcode and, if I?m reading the definitions of ?hashctrl? correctly, this appears to be the tri-state (2-bit) hashCode algorithm from Bacon et al. : https://doi.org/10.1007/3-540-47993-7_5. If that?s actually the case, it?d likely be good to include a citation in the code.

That is likely true. I can't access the paper, though, it asks me to pay 26€ ;-) I haven't read the paper, I've adopted the algorithm from talking to an OpenJ9 guy - OJ9 apparently uses a similar algorithm (haven't looked at their code, either). 

> It was a good idea 20 years ago, and remains a good idea. The only downside I know of is that you can exhaust memory extending objects in a moving GC, but there are ways to guard against that condition.

Right. It can't happen with sliding GCs, but copying GCs could theoretically run into this problem. It seems very very unlikely, but not impossible.

> The Compact Java Monitor approach is rather agonistic concerning the hashCode (and for that matter, anything else in the header, such as the class/klass information, and age bits). If you use the IBM 2-bit scheme, that?s fine, and if you need to displace it, that?s fine as well.
> 
> I hope to send out some results next week comparing the relative performance of a few potential ?synchronized? implementations.

I am a bit unsure about which direction to go. I also heard that there is work under way to remove JVM-side locking altogether, and use j.u.c instead, which would make the whole displaced-header problem go away. Getting rid of displaced headers would be a huge win. Otherwise we'll have to come up with a way to deal with it in concurrent GCs.

Thanks,
Roman

PR #15, opened 2021-09-27
Load Klass* from header in interpreter (x86)

rkennke:
This implements loading the compressed Klass* from the object header, instead of the Klass* field in the x86 interpreter. It does the fast-path (unlocked object) in assembly, and calls into the runtime to deal with locked objects.

Testing:
 - [x] tier1
 - [x] tier2
 - [x] hotspot_gc

### DIFF
--- a/src/hotspot/cpu/x86/macroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/macroAssembler_x86.cpp
@@ -4529,12 +4529,52 @@ void MacroAssembler::load_klass(Register dst, Register src, Register tmp) {
   assert_different_registers(src, tmp);
   assert_different_registers(dst, tmp);
 #ifdef _LP64
-  if (UseCompressedClassPointers) {
-    movl(dst, Address(src, oopDesc::klass_offset_in_bytes()));
-    decode_klass_not_null(dst, tmp);
-  } else
+  assert(UseCompressedClassPointers, "expect compressed class pointers");
+
+  Label slow, done;
+  movq(tmp, Address(src, oopDesc::mark_offset_in_bytes()));
+  // TODO: would be nice to use xorb instead, but it doesn't exist in our assembler. xorl zeroes the upper bits.
+  xorq(tmp, markWord::unlocked_value);
+  testb(tmp, 3/*markWord::lock_mask_in_place*/);
+  jccb(Assembler::notZero, slow);
+
+  movq(dst, tmp);
+  shrq(dst, markWord::klass_shift);
+  decode_klass_not_null(dst, tmp);
+  jmpb(done);
+  bind(slow);
+
+  if (dst != rax) {
+    push(rax);
+  }
+  push(rdi);
+  push(rsi);
+  push(rdx);
+  push(rcx);
+  push(r8);
+  push(r9);
+  push(r10);
+  push(r11);
+
+  MacroAssembler::call_VM_leaf(CAST_FROM_FN_PTR(address, oopDesc::load_klass_runtime), src);
+
+  pop(r11);
+  pop(r10);
+  pop(r9);
+  pop(r8);
+  pop(rcx);
+  pop(rdx);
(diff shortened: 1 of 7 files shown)


src/hotspot/cpu/x86/templateTable_x86.cpp
```diff
@@ -4184,12 +4181,12 @@ void TemplateTable::instanceof() {
   __ pop_ptr(rdx); // restore receiver
   __ verify_oop(rdx);
   Register tmp_load_klass = LP64_ONLY(rscratch1) NOT_LP64(noreg);
-  __ load_klass(rdx, rdx, tmp_load_klass);
-  __ jmpb(resolved);
+  __ load_klass(rdx, rdx, tmp_load_klass, false);
+  __ jmp(resolved);
```
shipilev, line 4185: Have you run into troubles with `jmpb` short branch here?

src/hotspot/cpu/x86/macroAssembler_x86.hpp
```diff
@@ -339,7 +339,7 @@ class MacroAssembler: public Assembler {
   void load_method_holder(Register holder, Register method);
 
   // oop manipulations
-  void load_klass(Register dst, Register src, Register tmp);
+  void load_klass(Register dst, Register src, Register tmp, bool null_check_src);
```
shipilev, line 342: I think if you define `bool null_check_src = false`, then the significant part of the changes would go away. It might not be a good idea when upstreaming this, but it would probably simplify the merges quite a bit.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -4538,16 +4538,64 @@ void MacroAssembler::load_method_holder(Register holder, Register method) {
...
+  assert(UseCompressedClassPointers, "expect compressed class pointers");
+
+  Label slow, done;
+  if (null_check_src) {
+    null_check(src, oopDesc::mark_offset_in_bytes());
+  }
+  movq(tmp, Address(src, oopDesc::mark_offset_in_bytes()));
+  // NOTE: While it would seem nice to use xorb instead (for which we don't have an encoding in our assembler),
+  // the encoding for xorq uses the signed version (0x81/6) of xor, which encodes as compact as xorb would,
+  // and does't make a difference performance-wise.
```
shipilev, line 4554: I can give you `xorb`, if you want, in a subsequent PR.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -4538,16 +4538,64 @@ void MacroAssembler::load_method_holder(Register holder, Register method) {
...
+    mov(dst, rax);
+    pop(rax);
+  }
+
+  bind(done);
+#else
+  if (null_check_src) {
+    null_check(src, oopDesc::klass_offset_in_bytes());
+  }
+  movptr(dst, Address(src, oopDesc::klass_offset_in_bytes()));
```
shipilev, line 4597: Do we not need to do any fixups for x86_32?

shipilev (approved):
Looks okay for the experimental code.

src/hotspot/cpu/x86/templateTable_x86.cpp
rkennke: Yes. The (subsequent) load_klass() increases the distance to resolved.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
rkennke: I actually have it, locally, but as comment says, it doesn't look any better in terms of code size or performance, so why bother?

src/hotspot/cpu/x86/templateTable_x86.cpp
shipilev: OK

PR #17, opened 2021-10-12
Fix use of Klass* in GC JFR reporting

rkennke:
Some JFR related code paths in GC access the Klass* of old objects that has already been overidden by the forwarding pointer. We can easily use the Klass* that we fetched earlier instead.

Testing:
 - [x] tier1
 - [x] tier2
 - [x] jdk/jfr (some of which still fails, see subsequent PRs)

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Change must be properly reviewed

### Reviewers
 * [Aleksey Shipilev](https://openjdk.java.net/census#shade) (@shipilev - Committer) ⚠️ Review applies to 1bf5407ffabc792b908a404e2d8543e5cb6b0507

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/17/head:pull/17` \
`$ git checkout pull/17`

Update a local copy of the PR: \
`$ git checkout pull/17` \
`$ git pull https://git.openjdk.java.net/lilliput pull/17/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 17`

View PR using the GUI difftool: \
`$ git pr show -t 17`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.java.net/lilliput/pull/17.diff">https://git.openjdk.java.net/lilliput/pull/17.diff</a>

</details>

PR #18, opened 2021-10-12
Use hashtable for obj->edge mapping in JFR, instead of mark-word

rkennke:
JFR overrides the mark-word to (temporarily) store a mapping from object to Edge*. This disturbs the Klass* that we need while tracing for leaks and for later emitting object information. Let's use a hashtable for this, instead, and leave the upper half of the mark-word alone. (The lower half is still used for marking.)

Testing:
 - [x] tier1
 - [x] tier2
 - [x] jdk/jfr (together with #17 most tests pass, needs one more follow-up test fix)

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/18/head:pull/18` \
`$ git checkout pull/18`

`$ git checkout pull/18` \
`$ git pull https://git.openjdk.java.net/lilliput pull/18/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 18`

`$ git pr show -t 18`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/18.diff">https://git.openjdk.java.net/lilliput/pull/18.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/jfr/leakprofiler/chains/edgeStore.hpp
+++ b/src/hotspot/share/jfr/leakprofiler/chains/edgeStore.hpp
@@ -57,9 +57,36 @@ class StoredEdge : public Edge {
   }
 };
 
+class ObjEdgeEntry {
+private:
+  oop _obj;
+  Edge* _edge;
+public:
+  ObjEdgeEntry(oop obj, Edge* edge) : _obj(obj), _edge(edge) {}
+  oop obj() const { return _obj; }
+  Edge* edge() const { return _edge; }
+};
+
+class ObjEdgeTableCallback : public JfrCHeapObj  {
+public:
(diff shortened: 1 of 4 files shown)


src/hotspot/share/jfr/leakprofiler/chains/edgeStore.hpp
```diff
@@ -57,9 +57,36 @@ class StoredEdge : public Edge {
   }
 };
 
+class ObjEdgeEntry {
+private:
+  oop _obj;
+  Edge* _edge;
```
shipilev, line 63: Should be `const`, I think.

rkennke:
I'm withdrawing this in favor of #21 which is mostly openjdk/jdk/pull/5918 as implemented by @mgronlun

PR #20, opened 2021-10-12
Load Klass from header, C1

rkennke:
This implements loading the Klass* from object header (instead of dedicated Klass* field) in C1 generated code.

Testing:
 - [x] tier1 (x86_64,x86_32, aarch64)
 - [x] tier2 (x86_64,x86_32, aarch64)
 - [x] tier3 (x86_64,x86_32, aarch64)
 - [ ] tier4 (x86_64,x86_32, aarch64)

---------
### Progress

### Reviewers
 * [Aleksey Shipilev](https://openjdk.java.net/census#shade) (@shipilev - Committer) ⚠️ Review applies to bc5b45e9e0d36304efdb9fd2648fe8b95ed73f8d

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/20/head:pull/20` \
`$ git checkout pull/20`

`$ git checkout pull/20` \
`$ git pull https://git.openjdk.java.net/lilliput pull/20/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 20`

`$ git pr show -t 20`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/20.diff">https://git.openjdk.java.net/lilliput/pull/20.diff</a>

</details>


src/hotspot/cpu/x86/c1_CodeStubs_x86.cpp
```diff
@@ -300,6 +300,22 @@ void MonitorExitStub::emit_code(LIR_Assembler* ce) {
...
 
+void LoadKlassStub::emit_code(LIR_Assembler* ce) {
+  __ bind(_entry);
+  Register res = _result->as_register();
+  ce->store_parameter(_obj->as_register(), 0);
+  if (res != rax) {
+    // This preserves rax and allows it to be used as return-register,
+    // without messing with the stack.
+    __ xchgptr(rax, res);
+  }
```
shipilev, line 311: I am trying to wrap my head around this in early Monday :) Is `res` guaranteed to be callee-save? I.e. does this stashing work if runtime call clobbers `res`?

src/hotspot/cpu/x86/c1_LIRAssembler_x86.cpp
```diff
@@ -1257,11 +1257,7 @@ void LIR_Assembler::mem2reg(LIR_Opr src, LIR_Opr dest, BasicType type, LIR_Patch
       break;
 
     case T_ADDRESS:
-      if (UseCompressedClassPointers && addr->disp() == oopDesc::klass_offset_in_bytes()) {
-        __ movl(dest->as_register(), from_addr);
-      } else {
-        __ movptr(dest->as_register(), from_addr);
-      }
+      __ movptr(dest->as_register(), from_addr);
```
shipilev, line 1260: Can we / should we `assert(addr->disp() != oopDesc::klass_offset_in_bytes(), "sanity")` here and later? So that the non-`LoadKlass` loads would break early?

src/hotspot/cpu/x86/c1_Runtime1_x86.cpp
```diff
@@ -1111,7 +1111,19 @@ OopMapSet* Runtime1::generate_code_for(StubID id, StubAssembler* sasm) {
...
+#ifdef _LP64
+      {
+        StubFrame f(sasm, "load_klass", dont_gc_arguments);
+        sasm->save_live_registers_no_oop_map(true);
+        f.load_argument(0, c_rarg0); // obj
+        __ call_VM_leaf(CAST_FROM_FN_PTR(address, oopDesc::load_klass_runtime), c_rarg0);
+        sasm->restore_live_registers_except_rax(true);
+      }
+#else
+      __ should_not_reach_here();
```
shipilev, line 1124: So the intent here to generate x86_32 stub, but check it is not called ever, right? I wonder if we still need to do `StubFrame f(sasm, "load_klass", dont_gc_arguments);` even for a simple `__ should_not_reach_here();`...

src/hotspot/cpu/x86/c1_LIRAssembler_x86.cpp
rkennke: Not sure, tbh. The current implementation is quite brittle, and we are lucky that we have apparently never hit (emit) a load with offset 8 (or 4 on x86_32) which happens to be klass_offset_in_bytes().

src/hotspot/cpu/x86/c1_Runtime1_x86.cpp
rkennke: Perhaps not. Let me see if I can avoid this altogether.

theRealAph:
This all looks reasonable and straightforward.

src/hotspot/share/oops/oop.cpp
```diff
@@ -177,7 +177,10 @@ void* oopDesc::load_oop_raw(oop obj, int offset) {
 
 JRT_LEAF(Klass*, oopDesc::load_klass_runtime(oopDesc* o))
   assert(o != NULL, "null-check");
-  return oop(o)->klass();
+  oop obj = oop(o);
+  assert(oopDesc::is_oop(obj), "need a valid oop here: " PTR_FORMAT, p2i(o));
+  Klass* klass = obj->klass();
+  return klass;
```
shipilev, line 183:
```suggestion
  return obj->klass();
```

src/hotspot/cpu/x86/c1_LIRAssembler_x86.cpp
```diff
@@ -3525,18 +3525,27 @@ void LIR_Assembler::emit_load_klass(LIR_OpLoadKlass* op) {
   Register obj = op->obj()->as_pointer_register();
   Register result = op->result_opr()->as_pointer_register();
 
-  CodeEmitInfo* info = op->info();
-  if (info != NULL) {
-    add_debug_info_for_null_check_here(info);
-  }
```
shipilev, line 3531: Wait, so this thing would not be executed for `!_LP64` anymore?

src/hotspot/cpu/x86/c1_CodeStubs_x86.cpp
shipilev: OK, taking that one on faith :)

src/hotspot/cpu/x86/c1_Runtime1_x86.cpp
```diff
@@ -1111,7 +1111,17 @@ OopMapSet* Runtime1::generate_code_for(StubID id, StubAssembler* sasm) {
       }
 
       break;
-
+#ifdef _LP64
+    case load_klass_id:
+      {
+        StubFrame f(sasm, "load_klass", dont_gc_arguments);
+        sasm->save_live_registers_no_oop_map(true);
```
shipilev, line 1118: I presume calling `*_no_oop_map` here is intentional, and valid for `call_VM_leaf`? I.e. no safepoint expected here?

src/hotspot/cpu/ppc/c1_CodeStubs_ppc.cpp
```diff
@@ -319,6 +319,9 @@ void MonitorExitStub::emit_code(LIR_Assembler* ce) {
   __ b(_continuation);
 }
 
+void LoadKlassStub::emit_code(LIR_Assembler* ce) {
+  // Currently not needed.
```
shipilev, line 323:
```suggestion
  // Currently not needed.
  Unimplemented();
```

src/hotspot/cpu/arm/c1_CodeStubs_arm.cpp
```diff
@@ -243,6 +243,9 @@ void MonitorExitStub::emit_code(LIR_Assembler* ce) {
   __ b(_continuation);
 }
 
+void LoadKlassStub::emit_code(LIR_Assembler* ce) {
+  // Currently not needed.
```
shipilev, line 247:
```suggestion
  // Currently not needed.
  Unimplemented();
```

src/hotspot/cpu/s390/c1_CodeStubs_s390.cpp
```diff
@@ -270,6 +270,10 @@ void MonitorExitStub::emit_code(LIR_Assembler* ce) {
   __ branch_optimized(Assembler::bcondAlways, _continuation);
 }
 
+void LoadKlassStub::emit_code(LIR_Assembler* ce) {
+  // Currently not needed.
```
shipilev, line 274:
```suggestion
  // Currently not needed.
  Unimplemented();
```

shipilev (approved):
Looks fine for experimental code, some nits below.

src/hotspot/cpu/x86/c1_LIRAssembler_x86.cpp
rkennke: Indeed! Good catch! Will fix before push.

src/hotspot/cpu/x86/c1_Runtime1_x86.cpp
rkennke: Yes exactly, leaf calls don't require an oop map.

rkennke: And no, no safepoints expected.

PR #21, opened 2021-10-13
Preserve Klass* in leak profiler

rkennke:
TBD


shipilev (approved):
All right. This is cherry-pick of upstream change, right? Looks good then.

PR #25, opened 2021-10-19
Load Klass* without causing monitor inflation

rkennke:
Until now, loading the Klass* in Lilliput may cause monitor inflation, because we needed a way to safely load the mark-word in the face of concurrent stack-locking or inflation happening. However, this caused troubles with concurrent GCs, because they may attempt to inflate monitors on from-space objects while traversing the heap for relocation/evacuation. Also, this may be a performance nuisance.
It turns out that we don't have to fully inflate the monitor: we only need a partial inflation, up to where we install the transient INFLATING word to prevent concurrent threads from messing with the mark-word, and then read and return the mark-word safely, and swing back the original real mark word to let other threads continue.

Testing:
 - [x] tier1
 - [x] tier2
 - [x] tier3
 - [x] tier4
 - [ ] hotspot_gc

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/25/head:pull/25` \
`$ git checkout pull/25`

`$ git checkout pull/25` \
`$ git pull https://git.openjdk.java.net/lilliput pull/25/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 25`

`$ git pr show -t 25`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/25.diff">https://git.openjdk.java.net/lilliput/pull/25.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/runtime/synchronizer.cpp
+++ b/src/hotspot/share/runtime/synchronizer.cpp
@@ -740,6 +740,79 @@ static markWord read_stable_mark(oop obj) {
   }
 }
 
+// Safely load a mark word from an object, even with racing stack-locking or monitor inflation.
+// The protocol is a partial inflation-protocol: it installs INFLATING into the object's mark
+// word in order to prevent an stack-locks or inflations from interferring (or detect such
+// interference and retry), but then, instead of creating and installing a monitor, simply
+// read and return the real mark word.
+markWord ObjectSynchronizer::safe_load_mark(oop object) {
+  for (;;) {
+    const markWord mark = object->mark_acquire();
+
+    // The mark can be in one of the following states:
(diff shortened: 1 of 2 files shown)


shipilev (approved):
OK, I am approving this because `ObjectSynchronizer::stable_mark` is Lilliput-specific, and so changes there do not affect the upstream locking code (apart from Lilliput-specific paths).

shipilev:
By the way, if this resolves the concurrent GC problems, should GHA workaround be removed too?
 https://builds.shipilev.net/patch-openjdk-lilliput/.github/workflows/submit.yml.udiff.html

PR #28, opened 2021-11-17
Deflate monitors of dead objects before they become unreachable

rkennke:
I've seen occasional failures in G1, where the refinement thread tries to parse a heap region that has dead objects, and would sometimes see an object with a monitor that has already been deflated. And because deflation does not bother to restore the header of dead objects, it would reach to unknown memory and crash.

The fix is to restore the header of dead objects just before they become unreachable. This can be done in the closures used by WeakProcessor::weak_oops_do(), right before the weak root will be cleared.

Testing:
 - [x] tier1
 - [x] tier2
 - [x] tier3
 - [x] tier4

---------
### Progress

### Reviewers
 * [Aleksey Shipilev](https://openjdk.java.net/census#shade) (@shipilev - Committer) ⚠️ Review applies to 88a4392adf7da7c93be304064ab9b3eec893fdc6

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/28/head:pull/28` \
`$ git checkout pull/28`

`$ git checkout pull/28` \
`$ git pull https://git.openjdk.java.net/lilliput pull/28/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 28`

`$ git pr show -t 28`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/28.diff">https://git.openjdk.java.net/lilliput/pull/28.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shared/oopStorage.inline.hpp
+++ b/src/hotspot/share/gc/shared/oopStorage.inline.hpp
@@ -31,6 +31,7 @@
 #include "metaprogramming/conditional.hpp"
 #include "metaprogramming/isConst.hpp"
 #include "oops/oop.hpp"
+#include "runtime/objectMonitor.hpp"
 #include "runtime/safepoint.hpp"
 #include "utilities/align.hpp"
 #include "utilities/count_trailing_zeros.hpp"
@@ -262,6 +263,7 @@ public:
       if (_is_alive->do_object_b(v)) {
         result = _f(ptr);
       } else {
+        ObjectMonitor::maybe_deflate_dead(ptr);
         *ptr = NULL;            // Clear dead value.
       }
     }
--- a/src/hotspot/share/gc/shared/weakProcessor.inline.hpp
+++ b/src/hotspot/share/gc/shared/weakProcessor.inline.hpp
@@ -65,6 +65,7 @@ public:
       _keep_alive->do_oop(p);
       ++_live;
     } else {
+      ObjectMonitor::maybe_deflate_dead(p);
       *p = NULL;
       ++_new_dead;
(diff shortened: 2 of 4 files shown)


shipilev (commented):
Well, the fix looks okay for the explanation.

But I do wonder. Is this affecting mainline too? How's Lilliput different that it encounters this bug? Should G1 even touch dead objects at all? I'd assume the object contents (including mark word) are generally garbage at that point?

rkennke:
> Well, the fix looks okay for the explanation.
> 
> But I do wonder. Is this affecting mainline too? How's Lilliput different that it encounters this bug? Should G1 even touch dead objects at all? I'd assume the object contents (including mark word) are generally garbage at that point?

I don't think it affects mainline. It only happens during refinement, and in some situations it needs to parse heap regions that contain unreachable, but not yet reclaimed, objects, using size-based traversal. In mainline this means it only needs to access the Klass* field, which is still in place, but in Lilliput this needs to access the object header, which may still point to a (already deflated) object monitor, because monitor deflation does not bother to update dead object's headers. This is what this patch fixes. I don't think that mainline ever accesses dead objects' headers.

shipilev:
> In mainline this means it only needs to access the Klass* field, which is still in place, but in Lilliput this needs to access the object header

Ah! Now I understand why Lilliput is affected by this. Okay then. Consider adding a blurb near `ObjectMonitor::maybe_deflate_dead`, something like: "// We might access the dead object headers for parsable heap walk, make sure headers are in correct shape, e.g. monitors deflated".

shipilev (approved):

PR #29, opened 2021-11-25
Load Klass* from header, C2 implementation

rkennke:
This implements loading the compressed Klass* from the object header in C2.
This is quite a hack: normally we should change the load offset everywhere in C2 from 8 (current klass_offset_in_bytes) to 0 (the mark word offset). However, offset = 0 has different meaning in C2, it means no offset, and throws off all sorts of stuff in C2. Realistically, in the long run, we want to change to loading from offset 4 (upper half of the header), but we currently need to deal with locking, and thus load from offset 0.

I fake it: I pretend to load from offset 4 everywhere in C2 ideal graph, but then load from offset 0 and check the mark bits instead. We can't leave the offset in ideal graph at 8 because that is going to conflict with loads from first field as soon as we drop the Klass*.

We also need to ensure that C2 knows that the condition flags are clobbered (cr), and that src and dst are not the same register.

Testing:
 - [x] tier1 (x86_64, aarch64)
 - [x] tier2 (x86_64, aarch64)
 - [x] tier3 (x86_64, aarch64)

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/29/head:pull/29` \
`$ git checkout pull/29`

`$ git checkout pull/29` \
`$ git pull https://git.openjdk.java.net/lilliput pull/29/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 29`

`$ git pr show -t 29`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/29.diff">https://git.openjdk.java.net/lilliput/pull/29.diff</a>

</details>


rkennke:
Also, @rwestrel please review and comment. Unfortunately, you don't seem to be a Lilliput committer, so can't officially approve. We should probably change that.

rwestrel:
Have you considered doing this in the IR graph instead of in platform dependent code? Maybe in Compile::final_graph_reshaping_impl() converting LoadNKlass to the markword load + logic to extract the class. Or do you think there's not much value in that approach?
The current implementation is ok AFAICT.

rkennke:
> Have you considered doing this in the IR graph instead of in platform dependent code? Maybe in Compile::final_graph_reshaping_impl() converting LoadNKlass to the markword load + logic to extract the class. Or do you think there's not much value in that approach? The current implementation is ok AFAICT.

Yes. That seems much more complex, though, especially considering that this is most likely a temporary solution anyway. At least as far as I know, efforts are ongoing upstream to get rid of the locking stuff in the mark word altogether, at which point we no longer need the special handling of load-klass. However, this might well be several years away, and I wanted to build a prototype now, and we already have the assembly code to handle LoadNKlass, so I thought the path of least resistance would be to reuse that. It should not make a measurable difference perf-wise, I think.

rwestrel:
> It should not make a measurable difference perf-wise, I think.

Right.

src/hotspot/share/opto/macro.cpp
```diff
@@ -1672,7 +1672,7 @@ PhaseMacroExpand::initialize_object(AllocateNode* alloc,
     transform_later(mark_node);
   }
   rawmem = make_store(control, rawmem, object, oopDesc::mark_offset_in_bytes(), mark_node, TypeX_X->basic_type());
-
+  rawmem = make_store(control, rawmem, object, oopDesc::nklass_offset_in_bytes(), klass_node, T_METADATA);
   rawmem = make_store(control, rawmem, object, oopDesc::klass_offset_in_bytes(), klass_node, T_METADATA);
```
shipilev, line 1676: This looks patently weird. What do we store to, `nklass` or `klass`? Why both here?

shipilev (commented):
I have comments :)

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -3713,20 +3713,19 @@ void MacroAssembler::load_nklass(Register dst, Register src) {
   b(done);
 
   bind(slow);
-  enter();
```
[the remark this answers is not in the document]
rkennke: enter() saves/restores lr and rfp, and then overrides rfp with current sp. leave() reverts that. C2 doesn't seem to like messing with rfp, but we still need to preserve lr at this point, so that subroutine calls can find their way back. The stub does its own enter/leave around the slow-path call.

src/hotspot/share/opto/macro.cpp
rkennke: The Klass* will be eliminated in the subsequent PR. Until then, we have to keep both.

rkennke: I believe I can safely remove the store to the nklass_offset. The narrow Klass* in the header is already initialized by storing the initial mark word. I only put the store to nklass_offset there to set-up a memory edge to subsequent load-klass. However, the initializer should always emit a membar, which provides the required memory ordering between the initializing store of the narrowKlass* and subsequent loads.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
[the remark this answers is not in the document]
shipilev: OK.

src/hotspot/share/opto/matcher.cpp
```diff
@@ -2977,7 +2977,7 @@ void State::dump(int depth) {
     tty->print("   ");
   }
   tty->print("--N: ");
-  _leaf->dump();
+  _leaf->dump(3);
```
shipilev, line 2980: Leftover debugging crumb?

shipilev (approved):
This PR is suddenly much shorter, is it still correct? The change seems fine, though.

shipilev (approved):
All right, let's do this.

PR #32, opened 2021-12-17
Implement Shenandoah support

rkennke:
This implements support for the Shenandoah GC in Lilliput. The following areas require special treatment in order for Shenandoah to work:

### JVMTI/JFR

I replace the fwdptr-resolve in the object scan loop with a proper LRB to prevent exposing from-space objects to JVMTI runtime.

### Stack-locking

Accessing the header for the Klass* (and therefore, for size) requires a special protocol to ensure that we're not chasing a stack-locked displaced header that is about to be unlocked, and therefore access potential garbage memory. This is done in #25. However, in Shenandoah it is slightly more complicated, because we need to access the size of an object in from-space, and don't want to observe a stack-lock that is about to be unlocked by another thread. In particular, I am worried about the following scenario:
T1:
1. leaves the final-mark safepoint, while holding lock O
2. Starts concurrent evacuation of O (concurrent threads evacuation)
3. CAS fwdptr to header of O
4. Unlocks O

T2 (possibly GC thread):
1. Starts evacuation of O
2. Accesses size/Klass*/header of O *in from space*, observe stack-lock
3. CAS fwdptr to header of O

If context switches after step 2, then T2 loads a stack-lock, then T1 succeeds to evacuate *and* unlock O, and then T2 accesses a dangling stack-lock.

We can use the same protocol that we implemented in #25 to prevent this: whenever we access the header of a from-space object, CAS 0 (INFLATING) into the header to prevent progress by any other thread, while at the same time get a safe hold on the stack-lock (or neutral lock if other thread was faster). In order for this to work, we need to change the evacuation protocol such that it retries (in busy-loop) when it observes a 0.
Same goes for any code that loads the mark-word in from-space (not all that many places). For loading the mark-word, we need to extend the protocol a little to allow reaching through the forwarding pointer. This is GC specific for Shenandoah. That is why I put the implementation for this into ShenandoahObjectUtils, which mostly mirrors ObjectSynchronizer implementation to load the mark word, with additional from-space object handling (which would not be necessary in regular runtime accesses to the mark-word, outside of GC code).

### Monitors

Monitors exhibit a similar problem: when observing a monitor while accessing an object's header, the concurrent deflater thread might concurrently deflate that monitor, and our thread might access a dangling monitor pointer. For Java threads, this is already prevented by the deflating protocol:
- First the deflater thread fixes all monitor headers back to neutral. During this phase, it is ok to racily load a monitor header: the monitor is still there, and the displaced header is safe to access.
- All Java threads are rendezvous'ed.
- Deflater destroys all deflated monitors. At this point, all Java thread would see a neutral header, and cannot access the destroyed monitors anymore.

This protocol is already extended by #27 to also rendezvous GC threads. This only requires that concurrent GC threads participate in SuspendibleThreadSet. Shenandoah has already implemented this, but turned off by default. The remaining step for Shenandoah to safely access monitor headers is to enable Suspendible GC workers.

Testing:
 - [x] hotspot_gc_shenandoah (x86_64, x86_32, aarch64)
 - [x] tier1 +UseShenandoahGC  (x86_64, x86_32, aarch64)
 - [x] tier2 +UseShenandoahGC  (x86_64, x86_32, aarch64)
 - [x] tier3 +UseShenandoahGC (x86_64, x86_32)
 - [ ] tier4 +UseShenandoahGC
 - [x] specjvm

---------
### Progress

### Reviewers
 * [Zhengyu Gu](https://openjdk.java.net/census#zgu) (@zhengyu123 - Committer) ⚠️ Review applies to fb48066915a84d77b6fa58a506a726e69d38b16e

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/32/head:pull/32` \
`$ git checkout pull/32`

`$ git checkout pull/32` \
`$ git pull https://git.openjdk.java.net/lilliput pull/32/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 32`

`$ git pr show -t 32`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/32.diff">https://git.openjdk.java.net/lilliput/pull/32.diff</a>

</details>


src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -334,7 +334,7 @@
           "How many times to maximum attempt to flush SATB buffers at the " \
           "end of concurrent marking.")                                     \
                                                                             \
-  product(bool, ShenandoahSuspendibleWorkers, false, EXPERIMENTAL,          \
+  product(bool, ShenandoahSuspendibleWorkers, true, EXPERIMENTAL,           \
```
zhengyu123, line 337: Why this is needed?

zhengyu123 (changes requested):

src/hotspot/share/gc/shenandoah/shenandoahObjectUtils.inline.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+#include "gc/shenandoah/shenandoahHeap.inline.hpp"
+#include "gc/shenandoah/shenandoahObjectUtils.hpp"
+#include "oops/klass.hpp"
+#include "oops/markWord.inline.hpp"
+#include "oops/oop.inline.hpp"
+#include "runtime/objectMonitor.inline.hpp"
+#include "runtime/thread.hpp"
+
+// This is a variant of ObjectSynchronizer::safe_load_mark(), which does the same thing, but also
```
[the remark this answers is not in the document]
rkennke: Done

src/hotspot/share/gc/shenandoah/shenandoahForwarding.inline.hpp
```diff
@@ -81,12 +81,18 @@ inline oop ShenandoahForwarding::try_update_forwardee(oop obj, oop update) {
...
   }
 
-  markWord new_mark = markWord::encode_pointer_as_mark(update);
-  markWord prev_mark = obj->cas_set_mark(new_mark, old_mark, memory_order_conservative);
-  if (prev_mark == old_mark) {
-    return update;
-  } else {
-    return cast_to_oop(prev_mark.clear_lock_bits().to_pointer());
+  while (true) {
+    markWord new_mark = markWord::encode_pointer_as_mark(update);
```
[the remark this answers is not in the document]
rkennke: Done

src/hotspot/share/gc/shenandoah/shenandoahObjectUtils.inline.hpp
```diff
@@ -0,0 +1,141 @@
...
+// handles forwarded objects. This is intended to be used by concurrent evacuation only. No other
+// code is supposed to observe from-space objects.
+#ifdef _LP64
+markWord ShenandoahObjectUtils::stable_mark(oop obj) {
```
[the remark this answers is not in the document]
rkennke: No, unfortunately not. We need to resolve inside the loop in order to catch concurrently evacuated objects. I considered the alternative to put some kind of GC hook into the loop in ObjectSynchronizer::stable_mark() but decided against it in order to keep shared code cleaner.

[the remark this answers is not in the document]
zhengyu123: Okay.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
zhengyu123: Any comments on this? Why does it relevant?

src/hotspot/share/gc/shenandoah/shenandoahForwarding.inline.hpp
```diff
@@ -82,11 +82,17 @@ inline oop ShenandoahForwarding::try_update_forwardee(oop obj, oop update) {
...
-    return update;
-  } else {
-    return cast_to_oop(prev_mark.clear_lock_bits().to_pointer());
+  while (true) {
+    markWord prev_mark = obj->cas_set_mark(new_mark, old_mark, memory_order_conservative);
+    if (prev_mark == old_mark) {
+      return update;
+    } else if (prev_mark == markWord::INFLATING()) {
+      // This happens when we encounter a stack-locked object in from-space.
+      // Busy-wait for completion.
```
zhengyu123, line 91: Can you explain how this can happen? I think we can only inflate to-space object, e.g. apply LRB first.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
rkennke: Ah yes. We must coordinate with concurrent monitor deflation. Specifically, we must prevent monitors from being deflated while we try to get hold of the object mark word for fetching the object size during evacuation. The protocol is relatively simple: relevant threads must respond to safepoint and/or suspension requests.
See also: https://github.com/openjdk/lilliput/blob/47392f37f85747c0e82d21ee475ffa2935c17a0b/src/hotspot/share/runtime/synchronizer.cpp#L1590

zhengyu123: I meant why it is relevant to this PR? Should this be fixed upstream first?

src/hotspot/share/gc/shenandoah/shenandoahForwarding.inline.hpp
rkennke: The trouble is that we need to fetch the size of an object during evacuation *before* the object is guaranteed to be in to-space. If we encounter an object that is stack-locked there, we may race with another thread which could, in the meantime, succeed to evacuate and unlock the object, in which case the stack-lock that we fetched from the from-space object would be invalid. We prevent that by CASing 0 into the header (e.g. the INFLATING marker) - this prevents both evacuation and unlocking, but it requires that our evacuation CAS is aware of it and retries if that happens.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
rkennke: No, involving GCs in the deflation protocol is only really relevant in Lilliput. And this change does it for Shenandoah. I was thinking to remove the flag altogether - not participating in deflation in Lilliput is a correctness problem and we should not even allow that.

zhengyu123: I see.

src/hotspot/share/gc/shenandoah/shenandoahForwarding.inline.hpp
zhengyu123: Hmm... it hurts my head to think about if we can see this transition state in from-space object, e.g. `ShenandoahForwarding::is_forwarded()` can return bogus result? all the from-space reads have to be stable reads?

rkennke: is_forwarded() will always be good: 00 means stack-locked or inflating, 01 means unlocked, 10 means monitor and 11 means forwarded.
There are places where from-space header accesses (to get object size and/or Klass*) requires extra care: all those cases have been changed to use the new ShenandoahObjectUtils functions instead of the usual stuff, and this helper takes care of dealing with concurrent evacuation (i.e. the otherwise unexpected 11 state).

zhengyu123: Okay, let's consider following scenario:
1. worker 1 tries to evac obj, reads its header and it is not `marked`
2. worker 1 reads obj size and allocates to-space obj
3. worker 2 tried to evac the same obj, reads its header and it is still not `marked`
4. worker 2 tries to read obj size and temporary installs `INFLATING` mark word
5. worker 1 copies obj to to-space with `INFLATING` mark word (?)
6. worker 1 wins CAS, so to-space obj has `INFLATING' mark word?

rkennke: At 6, worker 1 cannot win while the from-space object has INFLATING in its header. It will always try to CAS with the real original header. That is why we need the loop in the code that installs the forwarding pointer to retry when encountering INFLATING.

zhengyu123: Line #[79](https://github.com/openjdk/lilliput/blob/master/src/hotspot/share/gc/shenandoah/shenandoahForwarding.inline.hpp#L79) is not a stable read, right? so it can see `INFLATING` markword. I don't see why CAS at Line#[85](https://github.com/openjdk/lilliput/blob/master/src/hotspot/share/gc/shenandoah/shenandoahForwarding.inline.hpp#L85) has to fail? it may return `prev_mark = INFLATING`, no?

src/hotspot/share/gc/shenandoah/shenandoahForwarding.inline.hpp
```diff
@@ -76,13 +76,19 @@ inline bool ShenandoahForwarding::is_forwarded(oop obj) {
...
+    if (old_mark.is_marked()) {
+      return cast_to_oop(old_mark.clear_lock_bits().to_pointer());
+    }
+    if (old_mark.is_being_inflated()) {
+      // Inflating - this can only happen when concurrent evacuating thread tries to read
+      // the header for fetching the object Klass*/size. Retry.
+      old_mark = obj->mark();
+      continue;
+    }
     markWord prev_mark = obj->cas_set_mark(new_mark, old_mark, memory_order_conservative);
```
zhengyu123, line 92: I think you have to make sure that `update` oop has `old_mark` markword, in case it got `INFLATING` markword during copying.

src/hotspot/share/gc/shenandoah/shenandoahForwarding.inline.hpp
```diff
@@ -76,17 +76,32 @@ inline bool ShenandoahForwarding::is_forwarded(oop obj) {
...
-    return cast_to_oop(prev_mark.clear_lock_bits().to_pointer());
+  while (true) {
+    markWord prev_mark = obj->cas_set_mark(new_mark, old_mark, memory_order_conservative);
+    if (prev_mark == old_mark) {
+      return update;
+    } else if (prev_mark == markWord::INFLATING()) {
+      // This happens when we encounter a stack-locked object in from-space.
+      // Busy-wait for completion.
+      SpinPause();
+      continue;
```
zhengyu123, line 100: nit: do not need `continue`, otherwise, LGTM

zhengyu123 (approved):
LGTM

rkennke:
Ok thanks Zhengyu!

/integrate

PR #34, opened 2022-01-18
Fix loading Klass* in C1 on AArch64

rkennke:
PR #20 was not quite correct on AArch64:
- We cannot push/pop r0 in the code stub, because we must not mess with SP there, otherwise the argument passing does not work.
- We should use far_call() and not far_jump(), so that the return comes back into the stub, and not to the continuation (or whereever LR points to).

Testing:
 - [x] tier1
 - [x] tier2
 - [x] tier3

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/34/head:pull/34` \
`$ git checkout pull/34`

`$ git checkout pull/34` \
`$ git pull https://git.openjdk.java.net/lilliput pull/34/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 34`

`$ git pr show -t 34`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/34.diff">https://git.openjdk.java.net/lilliput/pull/34.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/c1_CodeStubs_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/c1_CodeStubs_aarch64.cpp
@@ -258,12 +258,23 @@ void LoadKlassStub::emit_code(LIR_Assembler* ce) {
   Register res = _result->as_register();
   ce->store_parameter(_obj->as_register(), 0);
   if (res != r0) {
-    __ push(RegSet::of(r0), sp);
+    // Note: we cannot push/pop r0 around the call, because that
+    // would mess with the stack pointer sp, and we need that to
+    // remain intact for store_paramater/load_argument to work correctly.
+    // We swap r0 and res instead, which preserves current r0 in res.
+    // The preserved value is later saved and restored around the
+    // call in Runtime1::load_klass_id.
+    __ mov(rscratch1, r0);
+    __ mov(r0, res);
+    __ mov(res, rscratch1);
   }
-  __ far_jump(RuntimeAddress(Runtime1::entry_for(Runtime1::load_klass_id)));
+  __ far_call(RuntimeAddress(Runtime1::entry_for(Runtime1::load_klass_id)));
   if (res != r0) {
-    __ mov(res, r0);
-    __ pop(RegSet::of(r0), sp);
+    // Swap back r0 and res. This brings the call return value
+    // from r0 into res, and the preserved value in res back into r0.
+    __ mov(rscratch1, r0);
+    __ mov(r0, res);
+    __ mov(res, rscratch1);
   }
   __ b(_continuation);
 }


rkennke:
@shipilev @adinn or @aph could you have a look, please?

shipilev (commented):
Why this new stub is different, i.e. why can it return not in `r0`? I am looking at e.g. `NewObjectArrayStub::emit_code`, and it has `assert(_result->as_register() == r0, "result must in r0");`. In fact, I cannot see any stub call with `far_call` that saves the registers itself...

rkennke:
> Why this new stub is different, i.e. why can it return not in `r0`? I am looking at e.g. `NewObjectArrayStub::emit_code`, and it has `assert(_result->as_register() == r0, "result must in r0");`. In fact, I cannot see any stub call with `far_call` that saves the registers itself...

The most correct way to do it would be to tell C1 to allocate the result register in r0. I tried that. However, it doesn't work with the code here:
https://github.com/openjdk/lilliput/blob/master/src/hotspot/share/gc/shared/c1/barrierSetC1.cpp#L337

Mostly because there is control flow (that probably shouldn't be there) and that confuses the register allocator.

theRealAph:
This doesn't look right. There's no point saving anything in res, because it may well just get clobbered by `Runtime1::load_klass_id`. Like Aleksey, I can't see why forcing res into `r0` won't work.

rkennke:
> This doesn't look right. There's no point saving anything in res, because it may well just get clobbered by `Runtime1::load_klass_id`. Like Aleksey, I can't see why forcing res into `r0` won't work.

I agree, and I would prefer to register-allocate res in r0. I tried that, and it works in most cases, but unfortunately C1 fails to do that correctly in this code sequence:
https://github.com/openjdk/lilliput/blob/master/src/hotspot/share/gc/shared/c1/barrierSetC1.cpp#L337

The reason for that is (according to @rwestrel) that the code sequence contains control flow inside a single basic block (which is borderline illegal, and we kinda lucky that it doesn't blow up), and that confuses the register allocator and would end up clobbering some other value. Fixing this would require to rewrite that code sequence to either emit several basic blocks and stitch together the control flow correctly. I don't really want to go there :-)

What I'm doing instead is allocate res freely, and then shuffle registers accordingly in the backend. What I'm doing here is this:
1. If res is allocated in r0 we don't have a problem at all, and don't need to do anything.
2. Else, I swap r0 and res. This preserves the original contents of r0 in res, which will be preserved around the call. This frees r0 for use as argument and result register.
3. After the call, I swap r0 and res again: This brings the result into res, and restores the original value of r0 from res (which has been push/popped around the call).

theRealAph:
>     1. If res is allocated in r0 we don't have a problem at all, and don't need to do anything.
> 
>     2. Else, I swap r0 and res. This preserves the original contents of r0 in res, which will be preserved around the call. This frees r0 for use as argument and result register.

Why will res be preserved around the call? That's the part I don't get.

rkennke:
> > ```
> > 1. If res is allocated in r0 we don't have a problem at all, and don't need to do anything.
> > 
> > 2. Else, I swap r0 and res. This preserves the original contents of r0 in res, which will be preserved around the call. This frees r0 for use as argument and result register.
> > ```
> 
> Why will res be preserved around the call? That's the part I don't get.

Because I implemented it that way (just like most other runtime call stubs):
https://github.com/openjdk/lilliput/blob/a9ec867aea4d665a1fc85f51bdb7586a575fcdcf/src/hotspot/cpu/aarch64/c1_Runtime1_aarch64.cpp#L725

theRealAph:
Ahh, I see. OK, it's kinda horrible, but I guess it'll be fixed later.

rkennke:
> Ahh, I see. OK, it's kinda horrible,

It's C1 :-/

> but I guess it'll be fixed later.

Famous last words :-D

rkennke:
> Ahh, I see. OK, it's kinda horrible, but I guess it'll be fixed later.

@shipilev is that ok for you too?

shipilev (approved):
OK, after reading the discussion, I understand what is going on here. Approving, but flickering through the horror-awe-disgust-amuzement cycle real hard here. :)

PR #35, opened 2022-01-25
Load narrowKlass from runtime, decode in generated code

rkennke:
Current implementation of load_klass_runtime() call ensures stable header, loads the narrowKlass from it, decodes it, and returns the Klass* back to generated (interpreter, C1) code. However, C2 wants to load the narrowKlass and use it directly. Let's change current impl to not decode in runtime, but decode in generated code instead, and thus avoid a new entry points for  C2. This doesn't even make generated code larger -- we only need to move the decode_klass() into the common (fast & slow) path.

The change also contains some exlusions for 32bit builds - narrow_klass() doesn't make sense (and doesn't compile) on x86_32, and thus shouldn't be compiled.

Testing:
 - [x] tier1 (x86_64, aarch64)
 - [x] tier2 (x86_64, aarch64)
 - [x] tier2 (x86_64, aarch64)

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/35/head:pull/35` \
`$ git checkout pull/35`

`$ git checkout pull/35` \
`$ git pull https://git.openjdk.java.net/lilliput pull/35/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 35`

`$ git pr show -t 35`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/35.diff">https://git.openjdk.java.net/lilliput/pull/35.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/c1/c1_LIRGenerator.cpp
+++ b/src/hotspot/share/c1/c1_LIRGenerator.cpp
@@ -1234,7 +1234,6 @@ void LIRGenerator::do_isInstance(Intrinsic* x) {
 void LIRGenerator::load_klass(LIR_Opr obj, LIR_Opr klass, CodeEmitInfo* null_check_info) {
   CodeStub* slow_path = new LoadKlassStub(obj, klass);
   __ load_klass(obj, klass, null_check_info, slow_path);
-  __ branch_destination(slow_path->continuation());
 }
 
 // Example: object.getClass ()
--- a/src/hotspot/cpu/aarch64/c1_LIRAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/c1_LIRAssembler_aarch64.cpp
@@ -2606,9 +2606,9 @@ void LIR_Assembler::emit_load_klass(LIR_OpLoadKlass* op) {
   // Fast-path: shift and decode Klass*.
   __ mov(result, tmp);
   __ lsr(result, result, markWord::klass_shift);
-  __ decode_klass_not_null(result);
 
   __ bind(*op->stub()->continuation());
+  __ decode_klass_not_null(result);
 }
 
 void LIR_Assembler::emit_profile_call(LIR_OpProfileCall* op) {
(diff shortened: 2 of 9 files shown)


src/hotspot/share/c1/c1_LIRGenerator.cpp
```diff
@@ -1234,7 +1234,6 @@ void LIRGenerator::do_isInstance(Intrinsic* x) {
 void LIRGenerator::load_klass(LIR_Opr obj, LIR_Opr klass, CodeEmitInfo* null_check_info) {
   CodeStub* slow_path = new LoadKlassStub(obj, klass);
   __ load_klass(obj, klass, null_check_info, slow_path);
-  __ branch_destination(slow_path->continuation());
```
shipilev, line 1237: Why this one is removed? I'd expect stub's `jmp(continuation)` to get here?

shipilev (approved):
(Tentatively approving to make bots happy)

> However, C2 wants to load the narrowKlass and use it directly.

Really? Why? I thought the universal contract for compressed oops/klasses is "decode on access, never actually deal with compressed forms"?

rkennke:
> (Tentatively approving to make bots happy)
> 
> > However, C2 wants to load the narrowKlass and use it directly.
> 
> Really? Why? I thought the universal contract for compressed oops/klasses is "decode on access, never actually deal with compressed forms"?

AFAIU, a lot of load-klass happens in order to check type, and a common fast-path for this is to compare one type for equality with another (before doing subtype checks, also for primitive types). It is not necessary to decode the nklass for equality check.

shipilev:
> > Really? Why? I thought the universal contract for compressed oops/klasses is "decode on access, never actually deal with compressed forms"?
> 
> AFAIU, a lot of load-klass happens in order to check type, and a common fast-path for this is to compare one type for equality with another (before doing subtype checks, also for primitive types). It is not necessary to decode the nklass for equality check.

I see, alright then.

src/hotspot/share/c1/c1_LIRGenerator.cpp
rkennke: It has actually been bound twice (to the same effective location), the other one is in c1_LIRAssembler_XXX.cpp:

https://github.com/rkennke/lilliput/blob/ab8ed6331fb4eb2c5eb0e13499178f8ef67dc0f6/src/hotspot/cpu/x86/c1_LIRAssembler_x86.cpp#L3546

It is important now to do it in C1_LIRAssembler, because we want the decode_klass() after the common path.

rkennke:
Thanks!

/integrate

PR #36, opened 2022-01-26
Load narrowKlass from header, AArch64 assembler implementation

rkennke:
This implements MacroAssembler::load_klass() to load the (narrow)Klass* from object header. Just like the x86_64 implementation, it checks whether it can take the fast path (object unlocked -> load from upper 32bits of header), or else calls the runtime to get a stable header and load from that.

It adds a runtime call stub, which will also be used in the C2 implementation. It also adds nklass_offset_in_bytes() which will also be used in C2 impl. The part in generate_verify_oop() is a little nasty, I added a comment that explains what's going on.

Testing:
 - [x] tier1 (aarch64)
 - [x] tier2 (aarch64)
 - [x] tier3 (aarch64)

---------
### Progress

### Reviewers
 * [Andrew Haley](https://openjdk.java.net/census#aph) (@theRealAph - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/36/head:pull/36` \
`$ git checkout pull/36`

`$ git checkout pull/36` \
`$ git pull https://git.openjdk.java.net/lilliput pull/36/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 36`

`$ git pr show -t 36`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/36.diff">https://git.openjdk.java.net/lilliput/pull/36.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -3708,6 +3708,8 @@ void MacroAssembler::load_klass(Register dst, Register src) {
     push(RegSet::of(tmp), sp);
   }
   assert_different_registers(src, tmp);
+  assert_different_registers(src, rscratch1);
+  assert_different_registers(src, rscratch2);
 
   Label slow, done;
 
@@ -3718,8 +3720,7 @@ void MacroAssembler::load_klass(Register dst, Register src) {
   br(Assembler::NE, slow);
 
   // Fast-path: shift and decode Klass*.
-  mov(dst, tmp);
-  lsr(dst, dst, markWord::klass_shift);
+  lsr(dst, tmp, markWord::klass_shift);
   b(done);
 
   bind(slow);
@@ -3730,12 +3731,9 @@ void MacroAssembler::load_klass(Register dst, Register src) {
   }
   // We don't need to preserve r0 here, but we need to preserve rscratch1 and rescratch2,
   // because some users of load_klass() use them around the call.
-  push(RegSet::of(rscratch1, rscratch2), sp);
   mov(r0, src);
   assert(StubRoutines::load_nklass() != NULL, "Must have stub");
-  mov(rscratch1, StubRoutines::load_nklass());
-  blr(rscratch1);
-  pop(RegSet::of(rscratch1, rscratch2), sp);
+  far_call(RuntimeAddress(StubRoutines::load_nklass()));
   if (dst != r0) {
     mov(dst, r0);
     pop(RegSet::of(r0), sp);
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -3723,7 +3723,6 @@ void MacroAssembler::load_klass(Register dst, Register src) {
   b(done);
 
   bind(slow);
-  set_last_Java_frame(sp, rfp, lr, rscratch1);
   enter();
   // We need r0 as argument and return register for the call. Preserve it, if necessary.
   if (dst != r0) {
@@ -3732,18 +3731,16 @@ void MacroAssembler::load_klass(Register dst, Register src) {
   // We don't need to preserve r0 here, but we need to preserve rscratch1 and rescratch2,
   // because some users of load_klass() use them around the call.
   push(RegSet::of(rscratch1, rscratch2), sp);
-  push_call_clobbered_registers_except(RegSet::of(r0));
   mov(r0, src);
-  mov(rscratch1, CAST_FROM_FN_PTR(address, oopDesc::load_nklass_runtime));
+  assert(StubRoutines::load_nklass() != NULL, "Must have stub");
+  mov(rscratch1, StubRoutines::load_nklass());
   blr(rscratch1);
-  pop_call_clobbered_registers_except(RegSet::of(r0));
   pop(RegSet::of(rscratch1, rscratch2), sp);
   if (dst != r0) {
     mov(dst, r0);
     pop(RegSet::of(r0), sp);
   }
   leave();
-  reset_last_Java_frame(true);
 
   bind(done);
   decode_klass_not_null(dst);
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -3693,11 +3693,61 @@ void MacroAssembler::load_method_holder(Register holder, Register method) {
 }
 
 void MacroAssembler::load_klass(Register dst, Register src) {
-  if (UseCompressedClassPointers) {
-    ldrw(dst, Address(src, oopDesc::klass_offset_in_bytes()));
-    decode_klass_not_null(dst);
-  } else {
-    ldr(dst, Address(src, oopDesc::klass_offset_in_bytes()));
+  assert(UseCompressedClassPointers, "expects UseCompressedClassPointers");
+
+  // We can receive src and dst in the same register here, and they can also
+  // come in rscratch1 and rscratch2. Let's allocate an additional register
+  // here to preserve src across the fast-path.
+  Register tmp = dst;
+  if (src == dst) {
+    if (src == r0) {
+      tmp = r1;
+    } else {
+      tmp = r0;
+    }
+    push(RegSet::of(tmp), sp);
+  }
+  assert_different_registers(src, tmp);
+
+  Label slow, done;
+
+  // Check if we can take the (common) fast path, if obj is unlocked.
+  ldr(tmp, Address(src, oopDesc::mark_offset_in_bytes()));
+  eor(tmp, tmp, markWord::unlocked_value);
+  tst(tmp, markWord::lock_mask_in_place);
+  br(Assembler::NE, slow);
+
+  // Fast-path: shift and decode Klass*.
+  mov(dst, tmp);
+  lsr(dst, dst, markWord::klass_shift);
+  decode_klass_not_null(dst);
+  b(done);
+
+  bind(slow);
+  set_last_Java_frame(sp, rfp, lr, rscratch1);
+  enter();
+  // We need r0 as argument and return register for the call. Preserve it, if necessary.
+  if (dst != r0) {
+    push(RegSet::of(r0), sp);
+  }
+  // We don't need to preserve r0 here, but we need to preserve rscratch1 and rescratch2,
+  // because some users of load_klass() use them around the call.
+  push(RegSet::of(rscratch1, rscratch2), sp);
+  push_call_clobbered_registers_except(RegSet::of(r0));
+  mov(r0, src);
+  mov(rscratch1, CAST_FROM_FN_PTR(address, oopDesc::load_klass_runtime));
+  blr(rscratch1);
+  pop_call_clobbered_registers_except(RegSet::of(r0));
+  pop(RegSet::of(rscratch1, rscratch2), sp);
+  if (dst != r0) {
+    mov(dst, r0);
+    pop(RegSet::of(r0), sp);
+  }
+  leave();
+  reset_last_Java_frame(true);
+  bind(done);
+  if (src == dst) {
+    pop(RegSet::of(tmp), sp);
   }
 }
 
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -3695,7 +3695,7 @@ void MacroAssembler::load_method_holder(Register holder, Register method) {
 // Loads the obj's Klass* into dst.
 // src and dst must be distinct registers
 // Preserves all registers (incl src, rscratch1 and rscratch2), but clobbers condition flags
-void MacroAssembler::load_klass(Register dst, Register src) {
+void MacroAssembler::load_nklass(Register dst, Register src) {
   assert(UseCompressedClassPointers, "expects UseCompressedClassPointers");
 
   assert_different_registers(src, dst);
@@ -3728,6 +3728,10 @@ void MacroAssembler::load_klass(Register dst, Register src) {
   leave();
 
   bind(done);
+}
+
+void MacroAssembler::load_klass(Register dst, Register src) {
+  load_nklass(dst, src);
   decode_klass_not_null(dst);
 }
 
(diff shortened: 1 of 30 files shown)


src/hotspot/share/oops/oop.hpp
```diff
@@ -307,6 +307,9 @@ class oopDesc {
 
   // for code generation
   static int mark_offset_in_bytes()      { return offset_of(oopDesc, _mark); }
+#ifdef _LP64
+  static int nklass_offset_in_bytes()    { return mark_offset_in_bytes() + markWord::klass_shift / 8; }
```
shipilev, line 311: Ew. Need to assert that `markWord::klass_shift` is actually a multiple of 8.

src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
```diff
@@ -584,8 +584,17 @@ class StubGenerator: public StubCodeGenerator {
...
+    // NOTE: We used to load the Klass* here, and compare that to zero.
+    // However, with current Lilliput implementation, that would require
+    // checking the locking bits and calling into the runtime, which
+    // clobbers the condition flags, which may be live around this call.
+    // OTOH, this is a simple NULL-check, and we can simply load the upper
+    // 32bit of the header as narrowKlass, and compare that to 0. The
+    // worst that can happen (rarely) is that the object is locked and
+    // we have lock pointer bits in the upper 32bits. We can't get a false
+    // negative.
+    __ ldrw(r0, Address(r0, oopDesc::nklass_offset_in_bytes()));  // get klass
```
shipilev, line 596: Something should assert that `oopDesc::nklass_offset_in_bytes()` does not break the alignment. Probably assert that `oopDesc::nklass_offset_in_bytes()` is a multiple of 4?

shipilev (commented):
I'll defer to Andrew to examine AArch64 code, but I have two quick comments.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -3693,11 +3693,59 @@ void MacroAssembler::load_method_holder(Register holder, Register method) {
...
+  }
+  assert_different_registers(src, tmp);
+
+  Label slow, done;
+
+  // Check if we can take the (common) fast path, if obj is unlocked.
+  ldr(tmp, Address(src, oopDesc::mark_offset_in_bytes()));
+  eor(tmp, tmp, markWord::unlocked_value);
+  tst(tmp, markWord::lock_mask_in_place);
+  br(Assembler::NE, slow);
```
theRealAph, line 3718:
```suggestion
  tbnz(tmp, exact_log2(markWord::lock_mask_in_place, slow);
```

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -3693,11 +3693,59 @@ void MacroAssembler::load_method_holder(Register holder, Register method) {
...
+
+  // Fast-path: shift and decode Klass*.
+  mov(dst, tmp);
+  lsr(dst, dst, markWord::klass_shift);
```
theRealAph, line 3722:
```suggestion
  lsr(dst, tmp, markWord::klass_shift);
```

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -3693,11 +3693,59 @@ void MacroAssembler::load_method_holder(Register holder, Register method) {
...
+  if (dst != r0) {
+    push(RegSet::of(r0), sp);
+  }
+  // We don't need to preserve r0 here, but we need to preserve rscratch1 and rescratch2,
+  // because some users of load_klass() use them around the call.
+  push(RegSet::of(rscratch1, rscratch2), sp);
+  mov(r0, src);
+  assert(StubRoutines::load_nklass() != NULL, "Must have stub");
+  mov(rscratch1, StubRoutines::load_nklass());
+  blr(rscratch1);
```
theRealAph, line 3737:
```suggestion
  far_call(RuntimeAddress(StubRoutines::load_nklass()), rscratch1);
```

src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
theRealAph: `ldrw()` will definitely complain if it isn't.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -3693,11 +3693,59 @@ void MacroAssembler::load_method_holder(Register holder, Register method) {
...
+
+  bind(slow);
+  enter();
+  // We need r0 as argument and return register for the call. Preserve it, if necessary.
+  if (dst != r0) {
+    push(RegSet::of(r0), sp);
+  }
+  // We don't need to preserve r0 here, but we need to preserve rscratch1 and rescratch2,
+  // because some users of load_klass() use them around the call.
+  push(RegSet::of(rscratch1, rscratch2), sp);
```
theRealAph, line 3733: This is all very weird, and looks confusing. I think it might be safe to force the callers of `load_klass()` to provide a temp.

src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
```diff
@@ -6595,6 +6605,27 @@ class StubGenerator: public StubCodeGenerator {
...
+    __ align(CodeEntryAlignment);
+    StubCodeMark mark(this, "StubRoutines", "load_nklass");
+
+    address start = __ pc();
+
+    __ set_last_Java_frame(sp, rfp, lr, rscratch1);
+    __ enter();
+    __ push_call_clobbered_registers_except(RegSet::of(r0));
+    __ call_VM_leaf(CAST_FROM_FN_PTR(address, oopDesc::load_nklass_runtime), 1);
+    __ pop_call_clobbered_registers_except(RegSet::of(r0));
```
theRealAph, line 6621:
```suggestion
    __ pop_call_clobbered_registers_except(r0);
```

src/hotspot/share/oops/oop.hpp
```diff
@@ -307,6 +307,12 @@ class oopDesc {
 
   // for code generation
   static int mark_offset_in_bytes()      { return offset_of(oopDesc, _mark); }
+#ifdef _LP64
+  static int nklass_offset_in_bytes()    {
+    assert(markWord::klass_shift % 8 == 0, "klass_shift must be multiple of 8");
```
theRealAph, line 312:
```suggestion
    STATIC_ASSERT(markWord::klass_shift % 8 == 0, "klass_shift must be multiple of 8");
```

src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
```diff
@@ -6595,6 +6605,27 @@ class StubGenerator: public StubCodeGenerator {
...
+  // Clobbers rscratch1
+  address generate_load_nklass() {
+    __ align(CodeEntryAlignment);
+    StubCodeMark mark(this, "StubRoutines", "load_nklass");
+
+    address start = __ pc();
+
+    __ set_last_Java_frame(sp, rfp, lr, rscratch1);
+    __ enter();
+    __ push_call_clobbered_registers_except(RegSet::of(r0));
```
theRealAph, line 6619:
```suggestion
    __ push_call_clobbered_registers_except(r0);
```
... everywhere.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
theRealAph: Oh, sorry. Right you are.

src/hotspot/share/oops/oop.hpp
rkennke: Done that.

src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
rkennke: Done that.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
rkennke: Done.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
rkennke: Done (without the rscratch1 arg, which is the default anyway).

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
rkennke: I checked again, and it actually looks like we don't need to preserve rscratch1 and rscratch2. I added asserts to verify that src register isn't one of the rscratch registers and would risk to get clobbered. It doesn't matter if dst is rscratch register.

src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
rkennke: Done.

src/hotspot/share/oops/oop.hpp
rkennke: Done.

src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
rkennke: Done.

rkennke:
@shipilev @theRealAph Thanks for reviewing! Please approve if you are ok with this updated PR.
Thanks, Roman

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -3693,11 +3693,57 @@ void MacroAssembler::load_method_holder(Register holder, Register method) {
...
+  b(done);
+
+  bind(slow);
+  enter();
+  // We need r0 as argument and return register for the call. Preserve it, if necessary.
+  if (dst != r0) {
+    push(RegSet::of(r0), sp);
+  }
+  // We don't need to preserve r0 here, but we need to preserve rscratch1 and rescratch2,
+  // because some users of load_klass() use them around the call.
```
theRealAph, line 3733: To what does this comment refer?

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -3693,11 +3693,57 @@ void MacroAssembler::load_method_holder(Register holder, Register method) {
...
+  assert_different_registers(src, rscratch1);
+  assert_different_registers(src, rscratch2);
+
+  Label slow, done;
+
+  // Check if we can take the (common) fast path, if obj is unlocked.
+  ldr(tmp, Address(src, oopDesc::mark_offset_in_bytes()));
+  eor(tmp, tmp, markWord::unlocked_value);
+  tst(tmp, markWord::lock_mask_in_place);
+  br(Assembler::NE, slow);
```
theRealAph, line 3720:
```suggestion
  andr(dst, tmp, markWord::lock_mask_in_place);
  cbnz(dst, Assembler::NE, slow);
```

src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
```diff
@@ -6595,6 +6605,27 @@ class StubGenerator: public StubCodeGenerator {
...
+  address generate_load_nklass() {
+    __ align(CodeEntryAlignment);
+    StubCodeMark mark(this, "StubRoutines", "load_nklass");
+
+    address start = __ pc();
+
+    __ set_last_Java_frame(sp, rfp, lr, rscratch1);
+    __ enter();
+    __ push_call_clobbered_registers_except(r0);
+    __ call_VM_leaf(CAST_FROM_FN_PTR(address, oopDesc::load_nklass_runtime), 1);
```
theRealAph, line 6620: I'm tempted to suggest saving flags here too. But I guess you've checked all the places that use `load_klass()`. At least the comment for `load_klass()` should advertise exactly what gets clobbered.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
rkennke: I believe that wouldn't work either. The fast path that is following this code requires that the upper 32 bits (the narrowKlass) of tmp remain intact: it shifts them to the right by 32bits and decodes the narrowKlass.

src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
rkennke: Hmm yeah, maybe. I found one instance where flags are live around this call, but could work around it:

https://github.com/rkennke/lilliput/blob/86c13e4dfc6c1e53c9c97fc686d299c4615fd28d/src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp#L587

Also, it's not only the stub that clobbers the condition flags, but the fast-path, too.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
theRealAph: > I believe that wouldn't work either. The fast path that is following this code requires that the upper 32 bits (the narrowKlass) of tmp remain intact: it shifts them to the right by 32bits and decodes the narrowKlass.

So what? tmp is untouched.

src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
theRealAph: Fast path is fixable.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
rkennke: > > I believe that wouldn't work either. The fast path that is following this code requires that the upper 32 bits (the narrowKlass) of tmp remain intact: it shifts them to the right by 32bits and decodes the narrowKlass.
> 
> So what? tmp is untouched.

We don't guarantee that tmp != dst. We only guarantee that tmp != src to avoid clobbering src.

theRealAph: OK, I give up. Please document 'clobbers flags' very clearly.

rkennke: > OK, I give up. Please document 'clobbers flags' very clearly.

I will try to make it that the caller ensures that src != dst, and also passes a tmp register, then we can avoid clobbering the flags register altogether and also avoid the register shuffling madness. Not sure if that's always feasable, though.

rkennke: Alright, ensuring src != dst is relatively easy and seems worth doing (makes fast-path faster). I tried using rscratch2 as tmp register and use the proposed andr/cbnz sequence, but apparently rscratch2 is live in at least one path, and I could not easily find it. I documented the register usage, and that the load_klass() clobbers condition flags. That seems good enough for now, especially considering that at some point in the future, it will all go away (when locking is not overlaying the Klass* stuff in the header anymore).
@theRealAph ok now?

rkennke:
I pushed more changes:
 - Split load_klass() and load_nklass() (we need the nklass version separately in a few places)
 - Cleanup remaining direct loads from klass_offset_in_bytes() in aarch64, replacing them with load_[n]klass()
 - Pass correct offset to null-checks that are preceding load_klass()
 - Trim code size estimate

@theRealAph and/or @shipilev please re-review? Thanks!

theRealAph (approved):
It looks good now.
Having said that, the convention is mostly that macros can use rscratchN registers freely internally, and that any caller needs to be aware of that. 
Here is the opposite of that convention: `load_nklass` carefully avoids clobbering `rscratch1` and `rscratch2` in order to make things easier for the caller. So, while this is something of a surprise for the maintainer, it's an elegant way to solve the problem.

rkennke:
> It looks good now. Having said that, the convention is mostly that macros can use rscratchN registers freely internally, and that any caller needs to be aware of that. Here is the opposite of that convention: `load_nklass` carefully avoids clobbering `rscratch1` and `rscratch2` in order to make things easier for the caller. So, while this is something of a surprise for the maintainer, it's an elegant way to solve the problem.

Thanks!
I believe the trouble is that load_klass() is used in some C1 generated code paths, and C1 register allocator doesn't ignore rscratch1 and rscratch2. I may be wrong, though.

theRealAph:
> Thanks! I believe the trouble is that load_klass() is used in some C1 generated code paths, and C1 register allocator doesn't ignore rscratch1 and rscratch2. I may be wrong, though.

I hope you are! - None of the register allocators can touch the scratch registers. Lots of hand-written code depends on that.

rkennke:
> > Thanks! I believe the trouble is that load_klass() is used in some C1 generated code paths, and C1 register allocator doesn't ignore rscratch1 and rscratch2. I may be wrong, though.
> 
> I hope you are! - None of the register allocators can touch the scratch registers. Lots of hand-written code depends on that.


> > Thanks! I believe the trouble is that load_klass() is used in some C1 generated code paths, and C1 register allocator doesn't ignore rscratch1 and rscratch2. I may be wrong, though.
> 
> I hope you are! - None of the register allocators can touch the scratch registers. Lots of hand-written code depends on that.

I had a look at the C1 register allocator, and could verify that I was wrong.

Thanks for reviewing and helping!

/integrate

PR #39, opened 2022-02-17
ObjectMonitor Storage

tstuefe:
Hi,

This prepares the way for an idea Roman had last year: to store OM references in a compressed form in the header, instead of as 64-bit pointer. Similar to narrow Klass pointers, we need an index or an offset into a memory region.

With this patch, OMs live in an array now. That new OM array is pre-reserved, gets committed on demand, has a freelist to manage released OMs. In a way, it is very similar to the old TSM solution before [JDK-8253064](https://bugs.openjdk.java.net/browse/JDK-8253064). Thanks a lot to @dcubed-ojdk for patiently explaining the details of the old solution to me([1], [2], [3]).

### Performance and memory use

I found that the renaissance "philosophers" benchmark is a good tool for measuring ObjectMonitor memory usage and performance. The benchmark, if you run with default VM options, does a lot of synchronization and creates millions of OMs. Running with a lower value for `-XX:MonitorUsedDeflationThreshold` diminishes the effect, and according to Dan [3] it would be a typical case for threshold reduction too.

This is not a typical use case. Typically we have only a few thousand OMs, and OM storage management does not matter that much. But we don't want huge performance drops in these outlier scenarios.

The first version of my patch was very naive and used a single global allocator and synchronized each access. Performance loss was brutal, about 15% compared to malloc.

I improved the patch to reduce contention:
- OMs are now preallocated in bulk on a per thread base (by default, 64, adjustable via `-XX:PreallocatedObjectMonitors`).
- OMs are released in bulk by preparing the OM freelist off-lock and only drawing the lock when appending the list to the central free list.

Again, all somewhat similar to the old TSM solution.

The improved version now has similar or even better performance and memory use than the stock VM, see below.

#### Measurements

Running renaissance philosphers benchmark with both Stock (Lilliput) VM and patched lilliput VM.

Options: `-XX:+UnlockDiagnosticVMOptions -Xmx2g -Xms2g -XX:NativeMemoryTracking=summary -XX:+PrintNMTStatistics -XX:+DumpVitalsAtExit`

We compare two allocators, one outside our control, in the libc. So I measure RSS, not Committed memory use, because I have no idea how much memory the libc commits in order to fulfill its mallocs. It may actually be a lot - to prevent contention, at least the glibc uses thread local arenas, which can get huge but are often mostly unused.

For the same reason I do not use `AlwaysPretouch`: we would pretouch committed mmaped memory in the patched version, but neither os::malloc() nor whatever overhead the libc produces would be touched. `AlwaysPretouch` would hence bias against stock. 

The performance numbers and RSS numbers wobbled a bit, but the following run is average (smaller numbers better):

##### Default run (`-XX:MonitorUsedDeflationThreshold=90`, `-XX:PreallocatedObjectMonitors=64`)

1) Stock

Benchmark result: 	4333,61537
Highest rss: 		3,7g

2) New ObjectMonitorStorage

Benchmark result: 	4122,14034 (+4.9%)
Highest rss: 		3,7g

##### Run with `-XX:MonitorUsedDeflationThreshold=50`

1) Stock

Benchmark result: 	4202,40876
Highest rss: 		rss=2,1g

2) New ObjectMonitorStorage

Benchmark result: 	4142,66361 (+1.4%)
Highest rss: 		1.9g (-9%)

##### Run with `-XX:PreallocatedObjectMonitors=1024`

2) New ObjectMonitorStorage

Benchmark result: 	4322,59806 (-2.8%)
Highest rss: 		3.5g (+66%)

#### Analysis

Patched version uses a bit less RSS and is 1..5% faster than unpatched version.

Enlarging the number of thread local preallocated OMs to 1024 was not so hot. Reduced contention comes at a high memory price and an actual performance loss too. There is an optimal point, maybe even the default of 64 is too high. I ran out of time though, but this is certainly a knob to optimize.

I suspect that if we go down this road, we'll need to spend more time optimizing memory and performance of OM storage. Even though my results are encouraging. But this is an area where a lot of tweaking happened upstream over time.

### Patch details:

The patch introduces the general-purpose classes `AddressStableArray`. A templatified array that is pre-reserved, is address stable, contiguous, committed on demand, keeps a freelist of released items. Code is well tested (see the new gtests) and can serve as building block for similar uses. E.g. Roman played with the idea of placing Thread into such an array too.

### What this patch does not:

This patch does not change the way OMs are stored in the markword. I had a quick glance, but its not trivial to find all places in generated code that read OMs from the mark word (e.g. `C2_MacroAssembler::rtm_inflated_locking`). I ran out of time and leave this for another day.

### Tests:

- GHAs
- SAP nightlies (scheduled)
- manual test on Linux X64, x86, aarch64, Windows x64

[1] https://mail.openjdk.java.net/pipermail/hotspot-runtime-dev/2022-January/053683.html
[2] https://mail.openjdk.java.net/pipermail/hotspot-runtime-dev/2022-February/053903.html
[3] https://mail.openjdk.java.net/pipermail/hotspot-runtime-dev/2022-March/054187.html

---------
### Progress
- [ ] Change must be properly reviewed

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/39/head:pull/39` \
`$ git checkout pull/39`

`$ git checkout pull/39` \
`$ git pull https://git.openjdk.java.net/lilliput pull/39/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 39`

`$ git pr show -t 39`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/39.diff">https://git.openjdk.java.net/lilliput/pull/39.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/runtime/globals.hpp
+++ b/src/hotspot/share/runtime/globals.hpp
@@ -2034,6 +2034,12 @@ const intx ObjectAlignmentInBytes = 8;
                                                                             \
   develop(bool, TraceOptimizedUpcallStubs, false,                           \
                 "Trace optimized upcall stub generation")                   \
+                                                                            \
+  product(uintx, MaxObjectMonitors, NOT_LP64(100000) LP64_ONLY(20 * M),     \
+                "Max. number of object monitors")                           \
+                                                                            \
+  product(uintx, PreallocatedObjectMonitors, NOT_LP64(8) LP64_ONLY(64),     \
+                "Max. thread local preallocated OMs")                       \
 
 // end of RUNTIME_FLAGS
 
--- a/src/hotspot/share/runtime/globals.hpp
+++ b/src/hotspot/share/runtime/globals.hpp
@@ -716,6 +716,13 @@ const intx ObjectAlignmentInBytes = 8;
           "before adjusting the in_use_list_ceiling up (0 is off).")        \
           range(0, max_uintx)                                               \
                                                                             \
+                                                                            \
+  product(uintx, MaxObjectMonitors, NOT_LP64(100000) LP64_ONLY(20 * M),     \
+           "Max. number of object monitors")                                \
+                                                                            \
+  product(uintx, PreallocatedObjectMonitors, NOT_LP64(8) LP64_ONLY(64),     \
+           "Max. thread local preallocated OMs")                            \
+                                                                            \
   product(intx, hashCode, 5, EXPERIMENTAL,                                  \
                "(Unstable) select hashCode generation algorithm")           \
                                                                             \
@@ -2034,12 +2041,7 @@ const intx ObjectAlignmentInBytes = 8;
                                                                             \
   develop(bool, TraceOptimizedUpcallStubs, false,                           \
                 "Trace optimized upcall stub generation")                   \
-                                                                            \
-  product(uintx, MaxObjectMonitors, NOT_LP64(100000) LP64_ONLY(20 * M),     \
-                "Max. number of object monitors")                           \
-                                                                            \
-  product(uintx, PreallocatedObjectMonitors, NOT_LP64(8) LP64_ONLY(64),     \
-                "Max. thread local preallocated OMs")                       \
+
 
 // end of RUNTIME_FLAGS
 
--- a/src/hotspot/share/runtime/globals.hpp
+++ b/src/hotspot/share/runtime/globals.hpp
@@ -716,9 +716,8 @@ const intx ObjectAlignmentInBytes = 8;
           "before adjusting the in_use_list_ceiling up (0 is off).")        \
           range(0, max_uintx)                                               \
                                                                             \
-                                                                            \
-  product(uintx, MaxObjectMonitors, NOT_LP64(100000) LP64_ONLY(20 * M),     \
-           "Max. number of object monitors")                                \
+  product(uintx, MonitorStorageSize, NOT_LP64(64*M) LP64_ONLY(4 * G),       \
+           "Size of object monitor store")                                  \
                                                                             \
   product(uintx, PreallocatedObjectMonitors, NOT_LP64(8) LP64_ONLY(64),     \
            "Max. thread local preallocated OMs")                            \
--- a/src/hotspot/share/runtime/mutexLocker.cpp
+++ b/src/hotspot/share/runtime/mutexLocker.cpp
@@ -247,7 +247,7 @@ void mutex_init() {
 
   def(Metaspace_lock               , PaddedMutex  , nosafepoint-3);
   def(MetaspaceCritical_lock       , PaddedMonitor, nosafepoint-1);
-  def(ObjectMonitorStorage_lock    , PaddedMutex,   nosafepoint);
+  def(ObjectMonitorStorage_lock    , PaddedMutex,   nosafepoint-3);
 
   def(Patching_lock                , PaddedMutex  , nosafepoint);      // used for safepointing and code patching.
   def(MonitorDeflation_lock        , PaddedMonitor, nosafepoint);      // used for monitor deflation thread operations
--- a/src/hotspot/share/runtime/mutexLocker.cpp
+++ b/src/hotspot/share/runtime/mutexLocker.cpp
@@ -153,6 +153,7 @@ Mutex*   UnregisteredClassesTable_lock= NULL;
 Mutex*   LambdaFormInvokers_lock      = NULL;
 #endif // INCLUDE_CDS
 Mutex*   Bootclasspath_lock           = NULL;
+Mutex*   ObjectMonitorStorage_lock    = NULL;
 
 #if INCLUDE_JVMCI
 Monitor* JVMCI_lock                   = NULL;
@@ -246,6 +247,7 @@ void mutex_init() {
 
   def(Metaspace_lock               , PaddedMutex  , nosafepoint-3);
   def(MetaspaceCritical_lock       , PaddedMonitor, nosafepoint-1);
+  def(ObjectMonitorStorage_lock    , PaddedMutex,   nosafepoint);
 
   def(Patching_lock                , PaddedMutex  , nosafepoint);      // used for safepointing and code patching.
   def(MonitorDeflation_lock        , PaddedMonitor, nosafepoint);      // used for monitor deflation thread operations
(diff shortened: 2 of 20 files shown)


dcubed-ojdk:
> But since AlwaysPreAlloc has no effect on malloced memory, it would benefit stock disproportionately.

Perhaps you meant `AlwaysPretouch` here instead of `AlwaysPreAlloc`.

dholmes-ora:
Drive by comment ...

Effectively reviving the old TSM-style allocation strategy does make me cringe somewhat. :( And without seeing how exactly this is to be used (including if the index goes out-of-range) makes it hard to suggest alternatives - though index into a (per-thread?) table of locked monitors seems one.

BTW shouldn't the freelist only be needed in JavaThread not Thread?

rkennke:
> Effectively reviving the old TSM-style allocation strategy does make me cringe somewhat. :( And without seeing how exactly this is to be used (including if the index goes out-of-range) makes it hard to suggest alternatives - though index into a (per-thread?) table of locked monitors seems one.

The motivation for this is to be able to compress the OM pointer to 32bit (or less), so that it fits into the lower part of the object header (or, later, the 32-bit-header). In order for this to work, we need a known address range for OMs into which we can allocate (aligned) OMs.

Roman

fisk (commented):
We worked very hard to get rid of TSM. I would cry a bit if this monster gets brought back from the dead.
I also don't know if this compression is sound. I have seen SpecJBB2015 create millions of ObjectMonitors. So it doesn't sound too far fetched to assume that a larger app, for example one with a TB sized heap, would run out of address space for ObjectMonitors, even if only a reasonable percentage of objects need a monitor.
I would very much prefer a solution that instead removes the ObjectMonitor pointer from the markWord completely, in favour of publishing the mapping with a (possibly optimized) hash table instead. Then you only need the hashCode in the markWord, TSM can stay dead, and it will scale better to large apps. Lookup time might be smaller, but that sounds like the better price to pay I think.

rkennke:
> We worked very hard to get rid of TSM. I would cry a bit if this monster gets brought back from the dead. I also don't know if this compression is sound. I have seen SpecJBB2015 create millions of ObjectMonitors. So it doesn't sound too far fetched to assume that a larger app, for example one with a TB sized heap, would run out of address space for ObjectMonitors, even if only a reasonable percentage of objects need a monitor. I would very much prefer a solution that instead removes the ObjectMonitor pointer from the markWord completely, in favour of publishing the mapping with a (possibly optimized) hash table instead. Then you only need the hashCode in the markWord, TSM can stay dead, and it will scale better to large apps. Lookup time might be smaller, but that sounds like the better price to pay I think.

Hmm, yeah.
However, this is not as simple as it seems, that global lookup table needs to be MT safe. Probably something based on concurrentHashTable.hpp, at least for a first cut (probably structurally similar to what I did in #11 for GC forwarding). Do you have an particular optimizations in mind for the use-case?

tstuefe:
> > We worked very hard to get rid of TSM. I would cry a bit if this monster gets brought back from the dead. I also don't know if this compression is sound. I have seen SpecJBB2015 create millions of ObjectMonitors. So it doesn't sound too far fetched to assume that a larger app, for example one with a TB sized heap, would run out of address space for ObjectMonitors, even if only a reasonable percentage of objects need a monitor. I would very much prefer a solution that instead removes the ObjectMonitor pointer from the markWord completely, in favour of publishing the mapping with a (possibly optimized) hash table instead. Then you only need the hashCode in the markWord, TSM can stay dead, and it will scale better to large apps. Lookup time might be smaller, but that sounds like the better price to pay I think.
> 
> Hmm, yeah. However, this is not as simple as it seems, that global lookup table needs to be MT safe. Probably something based on concurrentHashTable.hpp, at least for a first cut (probably structurally similar to what I did in #11 for GC forwarding). Do you have an particular optimizations in mind for the use-case?


Hi @fisk,  @rkennke,

I thought this through yesterday, but I am really snowed in, so sorry for not responding quickly.

I think it's not so easy. 

Let's say we add a Hashtable. We have concurrent adds and lookups. Lookups can happen in JIT, so we need a decoding mechanism in MacroAssembler, like we have with narrow Klass pointer decoding. But  but instead of a simple shift+add this would be a hashtable lookup. Depending on the hashtable this can be less or more complex (e.g. a closed hashing table may be somewhat simpler to code, but more memory consuming). But it is already more complex than shift+add. 

Atop of that, the concurrency aspect makes it really difficult. Maybe I'm just not seeing it.

Today in mainline, concurrency is handled silently by the libc. We have concurrent mallocs(). Libc handles this for us, at a certain silent expense (e.g. we pay RSS for that in the form of malloc arenas, or we have silent contention).

My OM store handles concurrency by bulk-allocating OMs and then a new allocation is usually just taking from a thread local freelist. But those bulk-allocated OMs are still inactive and have no association with an object. The association is done by just writing the OM pointer into the mark word, and the oop into the OM.

But a central hashmap would have to be maintained for every changed OM<->oop association. Concurrently to lookup, in compiled code.

I honestly think that such a solution would be more complex than a linear lookup table like my proposed OM store. Note that if you don't like the memory management aspect of it, we still could just have a linear pointer table, and keep OMs allocated in C-heap. That would reduce a bit of complexity at the cost of one more indirection and memory (both added libc overhead and the pointer table itself). The advantage would be that we would not have to pre-reserve a large memory range, less virtual memory usage.

About the "limit" aspect - lets say we have 32-bit narrow OM pointers, then we can have 4 billion OMs. Do you really see that as a limit? Arguably, at that point, we would have a lot of other problems as well. We would spend 1TB alone for OMs (with each OM being 256 bytes, if we align them to cache line size).

I'm still thinking. I'm sorry for causing pain, and I like reducing complexity much more than inflating it.

In my defense, I was hoping my solution (AddressStableArray and friends) could be reused in other places, e.g. as backing memory for Thread (or maybe even Klass, if we ever manage to make it homogeneously sized).

Cheers, Thomas

rkennke:
I realized that, at least in the long run, mapping oops to OMs in the header is not feasible. The problem is that it is not compatible with compact-hashcode. Compact hashcode can expand an object to include an additional field for storing the hashcode. This is done during GC. This means that the size of the original copy and the new copy of an object may differ. The size of an object is determined by looking at the object's header. If the header is displaced, e.g. into an OM, then we only have a single header for both copies of the object. This mean that we can not safely determine the object size during GC if we displace the header.

PR #40, opened 2022-02-17
Eliminate Klass* word

rkennke:
This change removes the dedicated Klass* word in the oop structure, at least in 64 bit builds. It has been usused, except for verification, since integration of #29.

The change makes regular objects smaller, the field layouter can now use the extra space and start laying out fields at offset 8 (instead of 12 or 16, depending on class-compression). However, arrays remain unchanged because they still store the array length in offset 8, and current array code expects the array elements to start at 64bit aligned address, even for smaller data types. I will address this in a subsequent change.

Much of the change is reverting the klass_offset -> nklass_offset change by #29, as promised. I also disabled s390x and ppc64le GHA builds, because they are broken by this change.

I problem-listed a number of SA tests. The SA is broken by this change because it needs to load the Klass* from a regular field. I tried to change it, but it looks very difficult: it doesn't have a way to safely access the object mark word in the face of concurrent activity by the locking subsystem. Fixing the SA is not very high priority at this point, it may ultimately not even be necessary when/if the locking subsystem doesn't use the mark word anymore (or at least, does no longer displace the header).

Testing:
 - [x] tier1 (x86_64, x86_32, aarch64)
 - [x] tier2 (x86_64, x86_32, aarch64)
 - [ ] tier3 (x86_64, x86_32, aarch64)

---------
### Progress

### Reviewers
 * [Aleksey Shipilev](https://openjdk.java.net/census#shade) (@shipilev - Committer) ⚠️ Review applies to fa9dcf62f0c265d4b490411303077e550638ce58

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/40/head:pull/40` \
`$ git checkout pull/40`

`$ git checkout pull/40` \
`$ git pull https://git.openjdk.java.net/lilliput pull/40/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 40`

`$ git pr show -t 40`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/40.diff">https://git.openjdk.java.net/lilliput/pull/40.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/parallel/psOldGen.cpp
+++ b/src/hotspot/share/gc/parallel/psOldGen.cpp
@@ -387,7 +387,7 @@ class VerifyObjectStartArrayClosure : public ObjectClosure {
     _start_array(start_array) { }
 
   virtual void do_object(oop obj) {
-    HeapWord* test_addr = cast_from_oop<HeapWord*>(obj) + 1;
+    HeapWord* test_addr = cast_from_oop<HeapWord*>(obj);
     guarantee(_start_array->object_start(test_addr) == cast_from_oop<HeapWord*>(obj), "ObjectStartArray cannot find start of object");
     guarantee(_start_array->is_block_allocated(cast_from_oop<HeapWord*>(obj)), "ObjectStartArray missing block allocation");
   }
--- a/test/langtools/ProblemList.txt
+++ b/test/langtools/ProblemList.txt
@@ -71,3 +71,6 @@ tools/sjavac/ClasspathDependencies.java                                         
 ###########################################################################
 #
 # jdeps
+
+
+jdk/jshell/ToolTabSnippetTest.java 1234567 generic-all
--- a/.github/workflows/submit.yml
+++ b/.github/workflows/submit.yml
@@ -421,14 +421,6 @@ jobs:
             debian-arch: armhf
             gnu-arch: arm
             gnu-flavor: eabihf
-          - flavor: hs s390x build only
-            flags: --enable-debug --disable-precompiled-headers
-            debian-arch: s390x
-            gnu-arch: s390x
-          - flavor: hs ppc64le build only
-            flags: --enable-debug --disable-precompiled-headers
-            debian-arch: ppc64el
-            gnu-arch: powerpc64le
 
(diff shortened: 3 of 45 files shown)


src/hotspot/share/gc/parallel/psOldGen.cpp
```diff
@@ -387,7 +387,7 @@ class VerifyObjectStartArrayClosure : public ObjectClosure {
     _start_array(start_array) { }
 
   virtual void do_object(oop obj) {
-    HeapWord* test_addr = cast_from_oop<HeapWord*>(obj) + 1;
+    HeapWord* test_addr = cast_from_oop<HeapWord*>(obj);
```
shipilev, line 390: A bit confused by this change, why?

test/langtools/ProblemList.txt
```diff
@@ -71,3 +71,6 @@ tools/sjavac/ClasspathDependencies.java
 ###########################################################################
 #
 # jdeps
+
+
+jdk/jshell/ToolTabSnippetTest.java 1234567 generic-all
```
shipilev, line 76: This still fails?

shipilev (approved):
Looks fine to me for the experimental code.

src/hotspot/share/gc/parallel/psOldGen.cpp
rkennke: The code there checks whether a routine can find the start of an object (by looking up in an offset table), starting from somewhere in the middle of an object. It allocates a minimal object, which is 2 words length, increases offset by 1, and starts searching from there. However, with this change, objects may be a single word, and increasing offset by one would find the next object instead. TBH, I wasn't quite sure what to do with that code.

test/langtools/ProblemList.txt
rkennke: Yeah, intermittendly, in GHA. It's not related to this change, but I wanted GHA green.

rkennke:
> Looks fine to me for the experimental code.

Thanks! I want to fix one more failing (because of changed object size) test in tier3 before I push it.

PR #44, opened 2022-03-08
Implement ZGC support

rkennke:
This change enables support for ZGC in Lilliput.

Similar to Shenandoah, the main trouble is concurrent relocation racing with stack-locking while trying to get hold of the header to figure out the object size. See #32 for details on this. Unfortunately, unlike Shenandoah, we cannot use the header as arbiter to prevent that race, because ZGC uses a forwarding table instead. For this reason I disabled stack-locking with ZGC, for now.

Other changes:
- Plant some STS usage, this is important to avoid racing header acess in GC workers with concurrent monitor deflation. Especially the ZRelocationTask and ZMarkTask don't seem to run under STS, and ZMarkRootTask has an STSJoiner in its constructor, but that would be called by the thread that drives the workers, not by the worker thread itself. I moved it into the worker. I also planted STS::yield() calls, so that safepoints (e.g. for monitor cleanup) can be reached in a timely manner. It may be useful to not do that after every object, but only after a certain number of strides, but that seems a minor and secondary problem.
- Removed an assert in ZBarrierSetC2 that seems wrong under Lilliput.

Testing:
 - [x] gc/z
 - [x] tier1 +UseZGC
 - [ ] tier2 +UseZGC
 - [ ] tier3 +UseZGC

---------
### Progress

### Reviewers
 * [Erik Österlund](https://openjdk.java.net/census#eosterlund) (@fisk - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/44/head:pull/44` \
`$ git checkout pull/44`

`$ git checkout pull/44` \
`$ git pull https://git.openjdk.java.net/lilliput pull/44/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 44`

`$ git pr show -t 44`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/44.diff">https://git.openjdk.java.net/lilliput/pull/44.diff</a>

</details>

### DIFF
--- a/.github/workflows/submit.yml
+++ b/.github/workflows/submit.yml
@@ -184,7 +184,7 @@ jobs:
       - name: Configure
         run: >
           bash configure
-          --with-jvm-features=-shenandoahgc,-zgc
+          --with-jvm-features=-shenandoahgc
           --with-conf-name=linux-x64
           ${{ matrix.flags }}
           --with-version-opt=${GITHUB_ACTOR}-${GITHUB_SHA}
@@ -540,7 +540,7 @@ jobs:
       - name: Configure
         run: >
           bash configure
-          --with-jvm-features=-shenandoahgc,-zgc
+          --with-jvm-features=-shenandoahgc
           --with-conf-name=linux-${{ matrix.gnu-arch }}-hotspot
           ${{ matrix.flags }}
           ${{ env.cross_flags }}
@@ -639,7 +639,7 @@ jobs:
       - name: Configure
         run: >
           bash configure
-          --with-jvm-features=-shenandoahgc,-zgc
+          --with-jvm-features=-shenandoahgc
           --with-conf-name=linux-x86
           --with-target-bits=32
           ${{ matrix.flags }}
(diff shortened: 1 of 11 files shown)


fisk (approved):
Looks good.

rkennke:
> Looks good.

Thanks, Erik.

However, I am back-paddling a little bit. Putting the STS in concurrent marking seems to break a number of tests. I am not quite sure why, though. OTOH, I believe it is necessary to do the STS during marking, because marking needs to access the Klass* (for object layout), and this needs be be safe wrt monitor deflation. I am investigating.

fisk:
> > Looks good.
> 
> Thanks, Erik.
> 
> However, I am back-paddling a little bit. Putting the STS in concurrent marking seems to break a number of tests. I am not quite sure why, though. OTOH, I believe it is necessary to do the STS during marking, because marking needs to access the Klass* (for object layout), and this needs be be safe wrt monitor deflation. I am investigating.

Generational ZGC already uses STS in these places (and a super special homegrown STS mechanism for relocation). I don't recall exactly what issues we ran into migrating the code to do that. :/

fisk (approved):
Looks good. Probably possible to get some deadlocks when in-place relocation is triggered from a load barrier in a safepoint, but we can deal with that separately.

rkennke:
Thanks, Erik!

/integrate

PR #45, opened 2022-03-23
JVMTI: Don't use object header for marking

rkennke:
JVMTI marks objects in order to track whether or not it has already visited objects during heap walking. This uses the usual GC marking bits in the object header. However, this proves to be confusing and brittle because some GCs also uses those header bits for marking and/or indicating forwarded objects. In particular, it becomes unreliable for Shenandoah GC to distinguish JVMTI marked objects from forwarded objects.

JVMTI should have no business in marking objects in their header. This change proposes to let JVMTI use its own (temporary) marking bitmap instead. This decouples JVMTI better from GCs.

Testing:
 - [x] tier1
 - [x] tier2
 - [ ] tier3

---------
### Progress
- [ ] Change must be properly reviewed

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/45/head:pull/45` \
`$ git checkout pull/45`

`$ git checkout pull/45` \
`$ git pull https://git.openjdk.java.net/lilliput pull/45/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 45`

`$ git pr show -t 45`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/45.diff">https://git.openjdk.java.net/lilliput/pull/45.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/prims/jvmtiTagMap.cpp
+++ b/src/hotspot/share/prims/jvmtiTagMap.cpp
@@ -28,8 +28,7 @@
 #include "classfile/symbolTable.hpp"
 #include "classfile/vmClasses.hpp"
 #include "classfile/vmSymbols.hpp"
-#include "gc/shared/collectedHeap.hpp"
-#include "gc/shared/markBitMap.inline.hpp"
+#include "gc/shared/objectMarker.hpp"
 #include "jvmtifiles/jvmtiEnv.hpp"
 #include "logging/log.hpp"
 #include "memory/allocation.inline.hpp"
@@ -1332,97 +1331,6 @@ jvmtiError JvmtiTagMap::get_objects_with_tags(const jlong* tags,
   return collector.result(count_ptr, object_result_ptr, tag_result_ptr);
 }
 
-
-// ObjectMarker is used to support the marking objects when walking the
-// heap.
-//
-// This implementation uses the existing mark bits in an object for
-// marking. Objects that are marked must later have their headers restored.
-// As most objects are unlocked and don't have their identity hash computed
-// we don't have to save their headers. Instead we save the headers that
-// are "interesting". Later when the headers are restored this implementation
-// restores all headers to their initial value and then restores the few
-// objects that had interesting headers.
-//
-// Future work: This implementation currently uses growable arrays to save
-// the oop and header of interesting objects. As an optimization we could
-// use the same technique as the GC and make use of the unused area
-// between top() and end().
-//
-
-// An ObjectClosure used to restore the mark bits of an object
-class RestoreMarksClosure : public ObjectClosure {
- public:
-  void do_object(oop o) {
-    if (o != NULL) {
-      markWord mark = o->mark();
-      if (mark.is_marked()) {
-        o->init_mark();
-      }
-    }
-  }
-};
-
-MarkBitMap ObjectMarker::_mark_bit_map;
-MemRegion  ObjectMarker::_bitmap_region;
-
-void ObjectMarker::initialize(MemRegion heap_region) {
-  new (&_mark_bit_map) MarkBitMap();
-  size_t bitmap_size = MarkBitMap::compute_size(heap_region.byte_size());
-  ReservedSpace bitmap(bitmap_size);
-  _bitmap_region = MemRegion((HeapWord*) bitmap.base(), bitmap.size() / HeapWordSize);
-  _mark_bit_map.initialize(heap_region, _bitmap_region);
-}
-
-// initialize ObjectMarker - prepares for object marking
-void ObjectMarker::init() {
-  assert(Thread::current()->is_VM_thread(), "must be VMThread");
-  assert(SafepointSynchronize::is_at_safepoint(), "must be at a safepoint");
-
-  // prepare heap for iteration
-  Universe::heap()->ensure_parsability(false);  // no need to retire TLABs
-
-  // create stacks for interesting headers
-  if (!os::commit_memory((char*)_bitmap_region.start(), _bitmap_region.byte_size(), false)) {
-    vm_exit_out_of_memory(_bitmap_region.byte_size(), OOM_MALLOC_ERROR,
-                          "Could not commit native memory for auxiliary marking bitmap for JVMTI object marking");
-  }
-  _mark_bit_map.clear();
-}
-
-// Object marking is done so restore object headers
-void ObjectMarker::done() {
-  if (!os::uncommit_memory((char*)_bitmap_region.start(), _bitmap_region.byte_size())) {
-    log_warning(gc)("Could not uncommit native memory for auxiliary marking bitmap for JVMTI object marking");
-  }
-}
-
-// mark an object
-inline void ObjectMarker::mark(oop o) {
-  assert(Universe::heap()->is_in(o), "sanity check");
-  assert(!o->mark().is_marked(), "should only mark an object once");
-  _mark_bit_map.mark(o);
-}
-
-// return true if object is marked
-inline bool ObjectMarker::visited(oop o) {
-  return _mark_bit_map.is_marked(o);
-}
-
-// Stack allocated class to help ensure that ObjectMarker is used
-// correctly. Constructor initializes ObjectMarker, destructor calls
-// ObjectMarker's done() function to restore object headers.
-class ObjectMarkerController : public StackObj {
- public:
-  ObjectMarkerController() {
-    ObjectMarker::init();
-  }
-  ~ObjectMarkerController() {
-    ObjectMarker::done();
-  }
-};
-
-
 // helper to map a jvmtiHeapReferenceKind to an old style jvmtiHeapRootKind
 // (not performance critical as only used for roots)
 static jvmtiHeapRootKind toJvmtiHeapRootKind(jvmtiHeapReferenceKind kind) {
@@ -1564,7 +1472,7 @@ class CallbackInvoker : AllStatic {
   // if the object hasn't been visited then push it onto the visit stack
   // so that it will be visited later
   static inline bool check_for_visit(oop obj) {
-    if (!ObjectMarker::visited(obj)) visit_stack()->push(obj);
+    if (!ObjectMarkerController::is_marked(obj)) visit_stack()->push(obj);
     return true;
   }
 
@@ -2851,8 +2759,8 @@ inline bool VM_HeapWalkOperation::collect_stack_roots() {
 //
 bool VM_HeapWalkOperation::visit(oop o) {
   // mark object as visited
-  assert(!ObjectMarker::visited(o), "can't visit same object more than once");
-  ObjectMarker::mark(o);
+  assert(!ObjectMarkerController::is_marked(o), "can't visit same object more than once");
+  ObjectMarkerController::mark(o);
 
   // instance
   if (o->is_instance()) {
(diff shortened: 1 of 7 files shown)


src/hotspot/share/prims/jvmtiTagMap.cpp
```diff
@@ -1362,28 +1363,16 @@ class RestoreMarksClosure : public ObjectClosure {
...
-};
-
-GrowableArray<oop>* ObjectMarker::_saved_oop_stack = NULL;
-GrowableArray<markWord>* ObjectMarker::_saved_mark_stack = NULL;
-bool ObjectMarker::_needs_reset = true;  // need to reset mark bits by default
+void ObjectMarker::initialize(MemRegion heap_region) {
+  new (&_mark_bit_map) MarkBitMap();
+  size_t bitmap_size = MarkBitMap::compute_size(heap_region.byte_size());
+  ReservedSpace bitmap(bitmap_size);
+  _bitmap_region = MemRegion((HeapWord*) bitmap.base(), bitmap.size() / HeapWordSize);
```
tstuefe, line 1373: I'd use BitsPerWord, not HeapWordSize. Potential conflict: MemRegion uses word size in HeapWord, but Bitmap ultimately expects a memory range in bm_word_t, and uses BitsPerWord internally.

Its a bit of a brain teaser. I would feel better if MarkBitMap::initialize() would assert that the bitmap region covers what it thinks it should cover. Currently it does not check the bitmap region size at all.

src/hotspot/share/prims/jvmtiTagMap.hpp
```diff
@@ -27,13 +27,28 @@
...
 class JvmtiTagMapTable;
 class JvmtiTagMapEntryClosure;
 
+// ObjectMarker provides the mark and visited functions
+class ObjectMarker : AllStatic {
+private:
+  static MarkBitMap _mark_bit_map;
+  static MemRegion  _bitmap_region;
+public:
+  static void initialize(MemRegion heap_region);
```
tstuefe, line 44: So, initialize() runs at VM startup, init() runs when the heap walk starts, if ever?

I'd rename init() to prepare_heapwalk or something similar. initialize vs init is not clear.

src/hotspot/share/prims/jvmtiTagMap.cpp
```diff
@@ -1394,55 +1383,30 @@ void ObjectMarker::init() {
   Universe::heap()->ensure_parsability(false);  // no need to retire TLABs
 
   // create stacks for interesting headers
-  _saved_mark_stack = new (ResourceObj::C_HEAP, mtServiceability) GrowableArray<markWord>(4000, mtServiceability);
-  _saved_oop_stack = new (ResourceObj::C_HEAP, mtServiceability) GrowableArray<oop>(4000, mtServiceability);
+  if (!os::commit_memory((char*)_bitmap_region.start(), _bitmap_region.byte_size(), false)) {
+    vm_exit_out_of_memory(_bitmap_region.byte_size(), OOM_MALLOC_ERROR,
+                          "Could not commit native memory for auxiliary marking bitmap for JVMTI object marking");
```
tstuefe, line 1388: Use os::commit_memory_or_exit ?

src/hotspot/share/prims/jvmtiTagMap.cpp
```diff
@@ -1394,55 +1383,30 @@ void ObjectMarker::init() {
...
-  RestoreMarksClosure blk;
-  if (needs_reset()) {
-    Universe::heap()->object_iterate(&blk);
-  } else {
-    // We don't need to reset mark bits on this call, but reset the
-    // flag to the default for the next call.
-    set_needs_reset(true);
+  if (!os::uncommit_memory((char*)_bitmap_region.start(), _bitmap_region.byte_size())) {
+    log_warning(gc)("Could not uncommit native memory for auxiliary marking bitmap for JVMTI object marking");
   }
```
tstuefe, line 1397: Clear bitmap?

tstuefe (commented):
Hi Roman,

Trying to understand.

You reserve a bitmap to cover the whole reserved heap space, with one bit per possible object location? So the memory use would be -Xmx / 12, right? (8 bits per byte, 3 bit shift)?

So I calculate ~90MB per GB address space. Am I thinking right?

This may hurt a bit for applications which use jvmti heap walk as OOM analysis tool. But I guess short term its okay.

Will in the course of walking every object be visited, so every bit be set eventually, or would the bitmap be more sparse? If the latter, we may reduce the footprint in the future by making a on-demand-commited bitmap, only committing pages with set bits. Would make clearing faster too.

Do we need a bitmap for the whole reserved range, would a committed range not simpler? But probably difficult to do if multiple committed regions exist.

Oh, but then more motivation in the future to support a sparse bitmap.

---

Comment before RestoreMarksClosure needs massaging.

Cheers, Thomas

src/hotspot/share/prims/jvmtiTagMap.cpp
tstuefe: Also, its not a OOM_MALLOC

tstuefe:
While thinking this through more fully, if the bitmap is used sparsely it won't hurt that much, since we don't touch a lot of memory, even if committed. It only hurts if you hit the commit limit.

tstuefe:
Just thinking, you probably don't need to clear the bitmap at all. os::commit_memory() remaps the space, so its a fresh mmap segment, and anonymous mmap segments are zero-initialized.

rkennke:
> You reserve a bitmap to cover the whole reserved heap space, with one bit per possible object location? So the memory use would be -Xmx / 12, right? (8 bits per byte, 3 bit shift)?

No. Each bit covers one word in Java heap (or more, if object alignment is higher). Bitmap size is 1/64 of heap size.
 
> So I calculate ~90MB per GB address space. Am I thinking right?

16MB/1GB address space.
 
> This may hurt a bit for applications which use jvmti heap walk as OOM analysis tool. But I guess short term its okay.

It should only be short-lived, as long as the heap walk takes.

> Will in the course of walking every object be visited, so every bit be set eventually, or would the bitmap be more sparse? If the latter, we may reduce the footprint in the future by making a on-demand-commited bitmap, only committing pages with set bits. Would make clearing faster too.

Yes, that would be an option. I assume it would be relatively sparse: objects are larger than single words, many regions are unused, e.g. to have headroom for GC, etc.

> Do we need a bitmap for the whole reserved range, would a committed range not simpler? But probably difficult to do if multiple committed regions exist.

Yes. We do that for example in Shenandoah GC: we only commit marking bitmaps for ranges where we have committed heap regions. But it requires better integration with GC. I have been thinking to move ObjectMarker into GC, accessible by a GC interface, and possibly implemented in a GC specific manner, and then GCs could optimize this, or even keep using object-header-marking if it doesn't cause problems with the GC (I believe it only really causes troubles with Shenandoah right now). Maybe this would be preferable overall. ?

Thanks,
Roman

tstuefe:
>>You reserve a bitmap to cover the whole reserved heap space, with one bit per possible object location? So the memory use would be -Xmx / 12, right? (8 bits per byte, 3 bit shift)?

>No. Each bit covers one word in Java heap (or more, if object alignment is higher). Bitmap size is 1/64 of heap size.

Right, thanks.

> > This may hurt a bit for applications which use jvmti heap walk as OOM analysis tool. But I guess short term its okay.
> 
> It should only be short-lived, as long as the heap walk takes.

Sure, but in OOM situations you may not have that memory. E.g. cloudfoundry jvmkill does a jvmti heapwalk before killing the VM due to OOM.

It is not an argument against your patch, especially since it just costs 1/64th. Just saying, it may degrade usefulness of jvmti heap walk. There are ways around it though, and possible optimizations later.

> 
> > Will in the course of walking every object be visited, so every bit be set eventually, or would the bitmap be more sparse? If the latter, we may reduce the footprint in the future by making a on-demand-commited bitmap, only committing pages with set bits. Would make clearing faster too.
> 
> Yes, that would be an option. I assume it would be relatively sparse: objects are larger than single words, many regions are unused, e.g. to have headroom for GC, etc.
> 
> > Do we need a bitmap for the whole reserved range, would a committed range not simpler? But probably difficult to do if multiple committed regions exist.
> 
> Yes. We do that for example in Shenandoah GC: we only commit marking bitmaps for ranges where we have committed heap regions. But it requires better integration with GC. I have been thinking to move ObjectMarker into GC, accessible by a GC interface, and possibly implemented in a GC specific manner, and then GCs could optimize this, or even keep using object-header-marking if it doesn't cause problems with the GC (I believe it only really causes troubles with Shenandoah right now). Maybe this would be preferable overall. ?

Up to you. I think it's fine as it is, and leaving this in jvmti has its charm too.

Personally, I probably would prefer a more generic solution, e.g. a bitmap which on the fly commits pages with set bits. Bitmap clear would be super simple, just uncommit the whole area. getbit() would return false for uncommitted areas. It is simpler in that you have no dependencies to GC or anything else, easier to reuse, saves more memory (since you only commit what is set, not what is committed). It would be an optimization one could do later though.

> Thanks, Roman

rkennke:
> > > This may hurt a bit for applications which use jvmti heap walk as OOM analysis tool. But I guess short term its okay.
> > 
> > 
> > It should only be short-lived, as long as the heap walk takes.
> 
> Sure, but in OOM situations you may not have that memory. E.g. cloudfoundry jvmkill does a jvmti heapwalk before killing the VM due to OOM.

Right. The current implementation also allocates native memory, in needs to preserve object headers of locked objects. It's probably very small, though.
 
> It is not an argument against your patch, especially since it just costs 1/64th. Just saying, it may degrade usefulness of jvmti heap walk. There are ways around it though, and possible optimizations later.

Hmm, yeah.

> > > Will in the course of walking every object be visited, so every bit be set eventually, or would the bitmap be more sparse? If the latter, we may reduce the footprint in the future by making a on-demand-commited bitmap, only committing pages with set bits. Would make clearing faster too.
> > 
> > 
> > Yes, that would be an option. I assume it would be relatively sparse: objects are larger than single words, many regions are unused, e.g. to have headroom for GC, etc.
> > > Do we need a bitmap for the whole reserved range, would a committed range not simpler? But probably difficult to do if multiple committed regions exist.
> > 
> > 
> > Yes. We do that for example in Shenandoah GC: we only commit marking bitmaps for ranges where we have committed heap regions. But it requires better integration with GC. I have been thinking to move ObjectMarker into GC, accessible by a GC interface, and possibly implemented in a GC specific manner, and then GCs could optimize this, or even keep using object-header-marking if it doesn't cause problems with the GC (I believe it only really causes troubles with Shenandoah right now). Maybe this would be preferable overall. ?
> 
> Up to you. I think it's fine as it is, and leaving this in jvmti has its charm too.

I would say, the marking bits in object header is GC concern, and managing a marking bitmap for heapwalk would also be a GC concern (it requires knowledge of heap layout, even more so if we wanted to implement your proposed optimizations).

> Personally, I probably would prefer a more generic solution, e.g. a bitmap which on the fly commits pages with set bits. Bitmap clear would be super simple, just uncommit the whole area. getbit() would return false for uncommitted areas. It is simpler in that you have no dependencies to GC or anything else, easier to reuse, saves more memory (since you only commit what is set, not what is committed). It would be an optimization one could do later though.

Yes. Let me implement a little abstraction in GC and a single implementation (which uses object header) for the start. I will do the bitmap-based heapwalk in the Shenandoah patch, and we can implement optimizations then.

Thank you!
Roman

rkennke:
I've reworked the change as follows:
- Added GC interface ObjectMarker and ObjectMarkerController by which JVMTI can interact with GC to do object marking.
- I provide two implementations: HeaderObjectMarker - which is 1:1 the existing implementation and BitmapObjectMarker which is the new implementation using a bitmap (still without any possible optimizations).
- Add a flag, mostly for testing, to switch implementations. Shenandoah may hardwire the flag later, or we drop the flag altogether and hardwire init_object_marker() in the CollectedHeap subclass.

This seems much cleaner than the previous implementation and also cleaner than the previous proposed change (no in-place constructor, no global state (except temporary global pointer to ObjectMarker impl), proper separation of concerns, etc).

coleenp:
Just a question.  Can you make this change for mainline?

rkennke:
> Just a question. Can you make this change for mainline?

ha! Yes sure I can, if you think there is interest. Maybe without the BitmapObjectMarker stuff, but possibly with some improvements to the oop/mark saving machinery.

rkennke:
> Just a question. Can you make this change for mainline?

See https://github.com/openjdk/jdk/pull/7964

rkennke:
Withdrawing this in favor of https://github.com/openjdk/jdk/pull/7964

PR #46, opened 2022-03-30
Simplify stable_mark() routine

rkennke:
The code in ObjectSynchronizer::stable_mark() is a bit over-complicated: it has a 'fast' path (which really isn't a fast-path, because the really-fast path is already in the various oopDesc::klass() methods), and for the slow-path loop dives into ObjectSynchronizer::safe_load_mark(). This change proposes to remove the stable_mark() method, and rename safe_load_mark() to stable_mark() instead.

I also propose a small improvement: we can use read_stable_mark() to deal with INFLATING mark properly, and handle the remaining cases in the loop: monitor and stack-locked are the most common - deal with them first. Stack-locked can safely access the mark when current thread is owning thread of the lock - avoid CASing INFLATING into the header in this case. Neutral and Marked cases go last, because they should not be very common (neutral is already caught in oopDesc::klass() variants, marked should only happen during full GCs by GC internal code).

Testing:
 - [x] tier1
 - [x] tier2
 - [x] tier3
 - [x] tier4

---------
### Progress

### Reviewers
 * [Zhengyu Gu](https://openjdk.java.net/census#zgu) (@zhengyu123 - no project role) ⚠️ Review applies to 36eebf490d37b4feb70bd9af8ca93e2d83b3a0d5
 * [Thomas Stuefe](https://openjdk.java.net/census#stuefe) (@tstuefe - Committer) ⚠️ Review applies to 36eebf490d37b4feb70bd9af8ca93e2d83b3a0d5

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/lilliput pull/46/head:pull/46` \
`$ git checkout pull/46`

`$ git checkout pull/46` \
`$ git pull https://git.openjdk.java.net/lilliput pull/46/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 46`

`$ git pr show -t 46`

</details>

<a href="https://git.openjdk.java.net/lilliput/pull/46.diff">https://git.openjdk.java.net/lilliput/pull/46.diff</a>

</details>


zhengyu123 (approved):
Looks reasonable to me.

src/hotspot/share/runtime/synchronizer.cpp
```diff
@@ -857,36 +852,13 @@ markWord ObjectSynchronizer::safe_load_mark(oop object) {
       return dmw;
```
tstuefe, line 852: Question, why is above guarantee a guarantee, not an assert? You want to be super sure here also in release builds? Do you plan to drop down to assert() eventually?

tstuefe (approved):
Hi,

looked through this and also the old version (I had missed that) and it looks fine. Interesting trick of temporary set the mark to "installing". I guess there is no code being surprised by the new state change (installing->locked)? Or could that happen before too?

Small remarks inline, but LGTM.

rkennke:
> Hi,
> 
> looked through this and also the old version (I had missed that) and it looks fine. Interesting trick of temporary set the mark to "installing". I guess there is no code being surprised by the new state change (installing->locked)? Or could that happen before too?

I haven't found any other code that would be surprised by this. That's probably because any such code would have to deal with it somehow already, when doing real monitor inflation. The trick has been introduced much earlier though, in #12 and later improved in #25, this PR only shuffles code around to be more compact and more efficient.

src/hotspot/share/oops/markWord.hpp
```diff
@@ -189,7 +189,7 @@ class markWord {
     return (BasicLock*) value();
   }
   bool has_monitor() const {
-    return ((value() & monitor_value) != 0);
```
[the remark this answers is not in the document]
rkennke: > So, the old version returned true for "monitor" (10) and for "marked" (11) both? Seems like a bug. Would it be worth fixing it independently, maybe upstream?

Yes. Maybe. All existing code seems to dodge this problem by either testing in a specific order, or not being affected by the marked state at all. It seems brittle and pointless to not do it correctly, though, so yeah, I'll propose it upstream.

src/hotspot/share/runtime/synchronizer.cpp
rkennke: TBH, dunno really. This code originates from ObjectSynchronizer::inflate() and we can probably turn it into an assert.

rkennke: I changed the guarantee to an assert, and will integrate as soon as GHA becomes green, unless you have objections.

rkennke:
One remaining GHA failure in Windows/x64 but looks unrelated:
jdk/jshell/HighlightUITest.java

Let's
/integrate

PR #51, opened 2022-06-01
Implement non-racy fast-locking

rkennke:
TBD Replace stack-locking with fast-locking scheme that is not racy on object header


shipilev:
Native ARM32 build fails, because `call_VM` there expects the argument to be in `R0`.

This helps:

```
diff --git a/src/hotspot/cpu/arm/interp_masm_arm.cpp b/src/hotspot/cpu/arm/interp_masm_arm.cpp
index 8480057ee99..1f8d4bb0f76 100644
--- a/src/hotspot/cpu/arm/interp_masm_arm.cpp
+++ b/src/hotspot/cpu/arm/interp_masm_arm.cpp
@@ -871,7 +871,8 @@ void InterpreterMacroAssembler::lock_object(Register Rlock) {
   ldr(Robj, Address(Rlock, obj_offset));
 
   // TODO: Implement fast-locking.
-  call_VM(noreg, CAST_FROM_FN_PTR(address, InterpreterRuntime::monitorenter), Robj);
+  mov(R0, Robj);
+  call_VM(noreg, CAST_FROM_FN_PTR(address, InterpreterRuntime::monitorenter), R0);
 }
 
 
@@ -890,7 +891,8 @@ void InterpreterMacroAssembler::unlock_object(Register Rlock) {
   ldr(Robj, Address(Rlock, obj_offset));
 
   // TODO: Implement fast-locking.
-  call_VM_leaf(CAST_FROM_FN_PTR(address, InterpreterRuntime::monitorexit), Robj);
+  mov(R0, Robj);
+  call_VM_leaf(CAST_FROM_FN_PTR(address, InterpreterRuntime::monitorexit), R0);
 }
```

rkennke:
> +  mov(R0, Robj);
> +  call_VM_leaf(CAST_FROM_FN_PTR(address, InterpreterRuntime::monitorexit), R0);

Thanks for testing it! I applied the suggested change.

src/hotspot/cpu/x86/c1_MacroAssembler_x86.cpp
```diff
@@ -60,72 +55,30 @@ int C1_MacroAssembler::lock_object(Register hdr, Register obj, Register disp_hdr
...
-  MacroAssembler::lock(); // must be immediately before cmpxchg!
-  cmpxchgptr(hdr, Address(obj, hdr_offset));
-  // if the object header was not pointing to the displaced header,
-  // we do unlocking via runtime call
-  jcc(Assembler::notEqual, slow_case);
-  // done
-  bind(done);
+
+  movptr(disp_hdr, Address(obj, hdr_offset));
+  andb(disp_hdr, ~0x3); // Clear lowest two bits. 8-bit AND preserves upper bits.
```
fisk, line 80: I see you added a new andb instruction so that you can clear two low order bits while preserving the others. It's worth noting that the immediates are sign extended. So I don't think you need to do that. For example you could do and with -4 of any signed immediate size to clear the low order 2 bits only.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -9461,5 +9439,55 @@ void MacroAssembler::get_thread(Register thread) {
...
+  cmpptr(tmp1, Address(thread, Thread::lock_stack_limit_offset()));
+  jcc(Assembler::greaterEqual, slow);
+
+  Register locked_hdr = tmp2->is_valid() ? tmp2 : tmp1;
+  // Now we attempt to take the fast-lock.
+  // hdr lowest two bits are either 00 or 01 here. Make it 01.
+  orptr(hdr, markWord::unlocked_value);
+  movptr(locked_hdr, hdr);
+  // Clear lowest two bits: we have 01 (see above), now flip the lowest to get 00.
+  xorptr(locked_hdr, markWord::unlocked_value);
```
fisk, line 9459: So you want to compute 1) the header but with low order bits are zero, and 2) the header but low order bits are 01.
I think I would compute the first one by using bitwise and with -4. Then I would compute the second by taking the first value + 1. The benefit is that you don't have to rely on the low order bits being either 00 or 01 in the header then, making it more future proof. Somebody might want to use 10 for something, for example.

fisk (commented):
Some initial comments in the assembly code for x86. I need to look more at this, but it's a start.

src/hotspot/cpu/x86/c1_MacroAssembler_x86.cpp
rkennke: Right. The downside is that the instruction encoding is larger (32 vs 16 bits, I believe). I don't think it matters much, though. I'll do what you suggest.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
rkennke: 10 is already used for monitors. Flipping the lowest bit would make it 11 - a value that should be impossible in this scenario, and which would make the CAS fail. However, what you suggest is clearer and doesn't rely on implicit knowledge, I'll do that, instead. Thank you!

dholmes-ora:
This sounds very much like the scheme that @robehn implemented for the Java object monitor project.

rkennke:
> This sounds very much like the scheme that @robehn implemented for the Java object monitor project.

Yeah, that is because we have been in exchange between @robehn, @fisk and myself about it. I also intend to upstream this change (minus the Lilliput-specific parts) soon, that will help Lilliput upstreaming and later on the JOM project.

dholmes-ora:
> Yeah, that is because we have been in exchange between @robehn, @fisk and myself about it. 

I expected that was the case but wanted to be clear there is prior work here.

> I also intend to upstream this change (minus the Lilliput-specific parts) soon, that will help Lilliput upstreaming and later on the JOM project.

I would need a lot of convincing that we should be doing anything upstream in this area "soon" given the current status of the two projects, but look forward to seeing such a proposal and its performance etc.

rkennke:
> > I also intend to upstream this change (minus the Lilliput-specific parts) soon, that will help Lilliput upstreaming and later on the JOM project.
> 
> I would need a lot of convincing that we should be doing anything upstream in this area "soon" given the current status of the two projects, but look forward to seeing such a proposal and its performance etc.

I don't know about JOM's status, because unfortunately the project is not public (otherwise I wouldn't have had to re-do it). Lilliput is at a point where I'm planning to start upstreaming it so that 64bit headers can make it into JDK21. And the new locking scheme would be one of the first prerequisite steps towards that goal. Performance in Lilliput: see above. I'd expect it to be neutral outside of Lilliput because there the benefit of faster load-Klass does not exist, and the benefit of faster i-hash is probably not significant.

src/hotspot/share/runtime/lockStack.hpp
```diff
@@ -0,0 +1,64 @@
...
+#ifndef SHARE_RUNTIME_LOCKSTACK_HPP
+#define SHARE_RUNTIME_LOCKSTACK_HPP
+
+#include "oops/oopsHierarchy.hpp"
+#include "utilities/globalDefinitions.hpp"
+#include "utilities/sizes.hpp"
+
+class Thread;
+class OopClosure;
+
```
tstuefe, line 34: I'd be curious if it would be good performance-wise to inline the object into Thread for the initial array size, and only heap-allocate if you go beyond that.
 
E.g.:

```
class LockStack {
   oop _initial[4];
   oop* _base, *_limit;
   LockStack() : _base(_initial), _limit(_initial + 4) {}
};
```
For the always-on-cost of 4 pointers in Thread you'd have one dereference hop less for most of the accesses and save 90% of mallocs.

Just a thought, one would have to measure.

tstuefe (commented):
Hi Roman,

Impressive work. I had a quick initial  glance at the changes. Some minor remarks inline.

One question: I tried to find out how you grow the LockStack array if we hit the limit while attempting to lock in compiled code. Looks to me you don't do that, or? Beyond recursive locking, that is another case where we inflate?

Also, I tried to understand in which cases LockStack::remove() is used. Obviously, if we recursively enter a lock we own, we inflate and remove the oop from our LockStack. But how does that work if another thread inflates a lock we have already fast locked?

Thanks, Thomas

rkennke:
> One question: I tried to find out how you grow the LockStack array if we hit the limit while attempting to lock in compiled code. Looks to me you don't do that, or? Beyond recursive locking, that is another case where we inflate?

I don't do that in compiled code. Instead, I'm checking for overflow, and call into the runtime for growing the array. No need to inflate the monitor, though.

> Also, I tried to understand in which cases LockStack::remove() is used. Obviously, if we recursively enter a lock we own, we inflate and remove the oop from our LockStack.

Right. Whenever a fast-lock is inflated, we remove the corresponding oop from the lock-stack.

> But how does that work if another thread inflates a lock we have already fast locked?

In this case, the other thread sets the owner of the new monitor to anonymous. As soon as we arrive at monitorexit, we know that the owner must be us, so we set that in the monitor, remove/pop the oop from the lock-stack, and perform a regular monitor exit to hand over the monitor to the other thread.

zhengyu123 (changes requested):

rkennke:
While preparing the change for upstream JDK (for comparison) I realized that I should trim and split it up into smaller independent parts. That should make it easier to review and also measure where performance changes actually come from.

rkennke:
> I would need a lot of convincing that we should be doing anything upstream in this area "soon" given the current status of the two projects, but look forward to seeing such a proposal and its performance etc.

See https://github.com/openjdk/jdk/pull/9680 for how the upstream change would look like (WIP, but working on x86_64 already). That's without Lilliput parts (load-Klass*) and without extra benefits (e.g. hashcode optimizations).

src/hotspot/cpu/x86/c1_MacroAssembler_x86.cpp
merykitty: Note that on x86 an instruction that read from a register that was written with a smaller-width instruction would result in register stall. For example in this occasion a later read on 32-bit of `disp_hdr` would be stalled as the last write is only 8-bit wide. As a result, it would be less efficient to use `andb` instead of `andptr` as you have fixed. Thanks.

rkennke: > Note that on x86 an instruction that read from a register that was written with a smaller-width instruction would result in register stall. For example in this occasion a later read on 32-bit of `disp_hdr` would be stalled as the last write is only 8-bit wide. As a result, it would be less efficient to use `andb` instead of `andptr` as you have fixed.
> 
> On a side note, they fix this issue with 32-bit write by making it an actual 64-bit write of the zero-extended value.
> 
> Thanks.

Ok, good to know. X86 is a funny platform to program assembly on. ;-)

rose00:
The discussion about `andb` vs `andq` or `andl` is interesting in its own right, and applies to lots of our assembly code, not just the current PR.  I put in a query to Sandhya on the Intel Java performance team to see if they have further advice; will share if there is anything.

rose00:
>  I put in a query to Sandhya on the Intel Java performance team to see if they have further advice; will share if there is anything.

sandhya.viswanathan@intel.com replied (thanks Sandhya!) as follows:

> From the “Intel Software Optimization Reference Manual” at:
> <https://www.intel.com/content/www/us/en/developer/articles/technical/intel-sdm.html>
> Section 3.5.2.3 of the manual indicates that a partial register stall occurs when an instruction refers to a register, portions of which were written previously by other instruction. This stall is much smaller on newer architectures. E.g. write to ah, bh, ch, or dh followed by a 2, 4, or 8 byte read of the same register can cause an overhead. 32-bit operation followed by a 64-bit read doesn’t incur that overhead.
> Excerpts from the manual:
> ----
> > General purpose registers can be accessed in granularities of bytes, words, doublewords; 64-bit mode also supports quadword granularity. Referencing a portion of a register is referred to as a partial register reference.
> > A partial register stall happens when an instruction refers to a register, portions of which were previously modified by other instructions. For example, partial register stalls occurs with a read to AX while previous instructions stored AL and AH, or a read to EAX while previous instruction modified AX.
> > The delay of a partial register stall is small in processors based on Intel Core microarchitecture, and in Pentium M processor (with CPUID signature family 6, model 13), Intel Core Solo, and Intel Core Duo processors. Pentium M processors (CPUID signature with family 6, model 9) and the P6 family incur a large penalty.
> > Note that in Intel 64 architecture, an update to the lower 32 bits of a 64 bit integer register is architecturally defined to zero extend the upper 32 bits. While this action may be logically viewed as a 32 bit update, it is really a 64 bit update (and therefore does not cause a partial stall).
> > Referencing partial registers frequently produces code sequences with either false or real dependencies.
> ----
> 
> So, it is recommended to use the same width operation on a GPR for up to 32bit data width. 32-bit operation followed by a 64-bit read is ok, so andl is ok.
> 
> For btc instruction REX.W indicates 64 bit. Between btcx and andx instruction andx has better throughput so better to use andx.


And here is my summary, FWIW.  I think something like this should go somewhere into `assembler.hpp`, since it is of general interest in the long term:

1. In order to change bits in the low 8 bits or low 16 bits of a 32-bit or 64-bit register, preserving the unaffected bits, use the full-sized operation, even if a narrower operation (e.g., movb, andb) seems nicer in some way.
2. If you don’t mind resetting the high 32 bits, a 32-bit operation on a 64-bit operand is fine also.
3. Failing to follow this advice has a significant pipeline penalty on older processors, and is also likely to have at least some cost on newer processors.  (Reference: Section 3.5.2.3, Intel Software Optimization Reference Manual.)
4. Even for single-bit updates, prefer and/or/xor to btr/bts/btc in registers.  The latter instructions, when locked on memory operands, may sometimes be useful for bitwise CAS-like operations.  However, full-word cmpxchg has more reliable performance, even for single-bit updates.
5. It is true that instruction cache pressure can degrade performance of an excessively sparse instruction workload, as whole.  But, because of deep prefetch, the effect on performance of memory encoding size of an individual instruction is usually insignificant.

robehn:
Funny, some of the code I "recognize" :)

There are some corner cases using this stack approach.
I have some new tests and some old ones which I'll run, if they seem happy I can start reviewing.

Can you please also run at least t4-t5 also.

rkennke:
> Funny, some of the code I "recognize" :)
> 
> There are some corner cases using this stack approach. I have some new tests and some old ones which I'll run, if they seem happy I can start reviewing.

Nice. But please use the isolated upstream change, instead. It received some fixed that are not present here.

https://github.com/openjdk/jdk/pull/9680

> Can you please run at least t4-t5 also.

Yes, will do. Thank you!

rkennke:
> Funny, some of the code I "recognize" :)
> 
> There are some corner cases using this stack approach. I have some new tests and some old ones which I'll run, if they seem happy I can start reviewing.
> 
> Can you please run at least t4-t5 also.

BTW, where do I find tier5+ tests? OpenJDK only seems to have test groups up to tier4.

rkennke:
I'm closing this for now. We will eventually adapt or merge from upstream change: https://bugs.openjdk.org/browse/JDK-8291555

PR #54, opened 2022-08-15
Replace stack-locking with non-racy fast-locking

rkennke:
This PR isolates the change to replace stack-locking with fast-locking (see #51) from the additional improvements to load-Klass* and hashcode and a few other minor places. It is derived from the proposed upstream change (https://github.com/openjdk/jdk/pull/9680) and has undergone a lot more additional testing.

This change replaces the current stack-locking implementation with a fast-locking scheme that retains the advantages of stack-locking (namely fast locking in uncontended code-paths), while avoiding the overload of the mark word. That overloading causes massive problems with Lilliput, because it means we have to check and deal with this situation. And because of the very racy nature, this turns out to be very complex and involved a variant of the inflation protocol to ensure that the object header is stable. 

What the original stack-locking does is basically to push a stack-lock onto the stack which consists only of the displaced header, and CAS a pointer to this stack location into the object header (the lowest two header bits being 00 indicate 'stack-locked'). The pointer into the stack can then be used to identify which thread currently owns the lock.

This change basically reverses stack-locking: It still CASes the lowest two header bits to 00 to indicate 'fast-locked' but does *not* overload the upper bits with a stack-pointer. Instead, it pushes the object-reference to a thread-local lock-stack. This is a new structure which is basically a small array of oop that is associated with each thread. Experience shows that this array typcially remains very small (3-5 elements). Using this lock stack, it is possible to query which threads owns which locks. Most importantly, the most common question 'does the current thread own me?' is very quickly answered by doing a quick scan of the array. More complex queries like 'which thread owns X?' are not performed in very performance-critical paths (usually in code like JVMTI or deadlock detection) where it is ok to do more complex operations.

In contrast to stack-locking, fast-locking does *not* support recursive locking (yet). When that happens, the fast-lock gets inflated to a full monitor. It is not clear if it is worth to add support for recursive fast-locking.

One trouble is that when a contending thread arrives at a fast-locked object, it must inflate the fast-lock to a full monitor. Normally, we need to know the current owning thread, and record that in the monitor, so that the contending thread can wait for the current owner to properly exit the monitor. However, fast-locking doesn't have this information. What we do instead is to record a special marker ANONYMOUS_OWNER. When the thread that currently holds the lock arrives at monitorexit, and observes ANONYMOUS_OWNER, it knows it must be itself, fixes the owner to be itself, and then properly exits the monitor, and thus handing over to the contending thread.

As an alternative, I considered to remove stack-locking altogether, and only use heavy monitors. In most workloads this did not show measurable regressions. However, in a few workloads, I have observed severe regressions. All of them have been using old synchronized Java collections (Vector, Stack), StringBuffer or similar code. But alas, such code exists, and we probably don't want to punish it if we can avoid it.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput pull/54/head:pull/54` \
`$ git checkout pull/54`

`$ git checkout pull/54` \
`$ git pull https://git.openjdk.org/lilliput pull/54/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 54`

`$ git pr show -t 54`

</details>

<a href="https://git.openjdk.org/lilliput/pull/54.diff">https://git.openjdk.org/lilliput/pull/54.diff</a>

</details>


src/hotspot/cpu/aarch64/aarch64.ad
```diff
@@ -4075,38 +4047,43 @@ encode %{
...
-      __ tst(oop, oop); // Set NE to indicate 'failure' -> take slow-path. We know that oop != 0.
+      __ fast_unlock(oop, tmp, box, disp_hdr, slow);
+
+      // Indicate success at cont.
+      __ cmp(oop, oop);
+      __ b(cont);
+
+      __ bind(slow);
     }
+    __ tst(oop, oop); // Set NE to indicate 'failure' -> take slow-path. We know that oop != 0.
```
theRealAph, line 4067:
```suggestion
    __ cmp(zr, (u1)1); // Set NE to indicate 'failure' -> take slow-path..
```

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -5953,3 +5953,49 @@ void MacroAssembler::double_move(VMRegPair src, VMRegPair dst, Register tmp) {
...
+// Registers:
+//  - obj: the object to be locked
+//  - hdr: the header, already loaded from obj, will be destroyed
+//  - t1, t2, t3: temporary registers, will be destroyed
+void MacroAssembler::fast_lock(Register obj, Register hdr, Register t1, Register t2, Register t3, Label& slow) {
+  // Check if we would have space on lock-stack for the object.
+  ldr(t1, Address(rthread, Thread::lock_stack_current_offset()));
+  ldr(t2, Address(rthread, Thread::lock_stack_limit_offset()));
+  cmp(t1, t2);
+  br(Assembler::GE, slow);
```
theRealAph, line 5968:
```suggestion
  br(Assembler::HS, slow);
```
Are these really 64-bit unsigned quantities, though?

rkennke:
Closing in favour of eventual merging from upstream.

PR #57, opened 2022-10-07
Simplify loading of Klass*

rkennke:
With the arrival of fast-locking, it is now possible to greatly simplify the loading of Klass* from object headers. We now only need to check if the object is locked by a monitor, and if so, fetch the real mark from the monitor. This is a safe and non-racy operation, because Java and GC threads coordinate with monitor deflation.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput pull/57/head:pull/57` \
`$ git checkout pull/57`

`$ git checkout pull/57` \
`$ git pull https://git.openjdk.org/lilliput pull/57/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 57`

`$ git pr show -t 57`

</details>

<a href="https://git.openjdk.org/lilliput/pull/57.diff">https://git.openjdk.org/lilliput/pull/57.diff</a>

</details>


src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -5091,12 +5091,11 @@ void MacroAssembler::load_nklass(Register dst, Register src) {
   // NOTE: While it would seem nice to use xorb instead (for which we don't have an encoding in our assembler),
```
forax, line 5091: Hi Roman, the comment above is now stale, i think you can remove it

rkennke:
Closing in favour of more up-to-date, complete and better performing #66

PR #63, opened 2023-01-18
8300573: [Lilliput] Properly set ZF on anon-check path; avoid some conditional branches

rkennke:
In AArch64's fast_unlock(), we need to check if the monitor owner is ANONYMOUS (only with Lilliput/fast-locking), and if so, call the runtime to fix this before exiting the monitor. I tried to be smart and used tbnz there, but it turns out that tbnz does *not* set the zero flag, but we do need the ZF set correctly on exit of fast_unlock, so that C2 generated code would actually call the slow path. Therefore we need to use the longer tst/br instructions instead.

The change also removes some conditional branches.

All of the changes have already been integrated into the upstream proposed PR: https://github.com/openjdk/jdk/pull/10907/

Testing:
 - [x] tier1
 - [x] tier2

---------
### Progress

### Issue
 * [JDK-8300573](https://bugs.openjdk.org/browse/JDK-8300573): [Lilliput] Properly set ZF on anon-check path; avoid some conditional branches

### Reviewers
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - Committer) ⚠️ Review applies to [c6726ed0](https://git.openjdk.org/lilliput/pull/63/files/c6726ed0703b657c311c0f5d70cc0ac2c641bd0c)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput pull/63/head:pull/63` \
`$ git checkout pull/63`

`$ git checkout pull/63` \
`$ git pull https://git.openjdk.org/lilliput pull/63/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 63`

`$ git pr show -t 63`

</details>

<a href="https://git.openjdk.org/lilliput/pull/63.diff">https://git.openjdk.org/lilliput/pull/63.diff</a>

</details>


shipilev (approved):
Uff. `cont` and `count` look cunningly similar. Anyway!

PR #65, opened 2023-01-20
8300735: [Lilliput] Adjust GHA and jcheck config

rkennke:
I'd like to exclude not-yet-supported platforms that are failing in GHA (ppc64le, risc64, s390x). Also, we should remove a stale submit.yml.
I would also like to adjust jcheck to require a JBS issue for each PR, as I have recently begun to do. This improves traceability and discoverability of Lilliput work in JBS.

---------
### Progress

### Issue
 * [JDK-8300735](https://bugs.openjdk.org/browse/JDK-8300735): [Lilliput] Adjust GHA and jcheck config

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput pull/65/head:pull/65` \
`$ git checkout pull/65`

`$ git checkout pull/65` \
`$ git pull https://git.openjdk.org/lilliput pull/65/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 65`

`$ git pr show -t 65`

</details>

<a href="https://git.openjdk.org/lilliput/pull/65.diff">https://git.openjdk.org/lilliput/pull/65.diff</a>

</details>


shipilev:
I think it would be enough to just trim the defaults here:

```
 default: "Linux additional (hotspot only), Linux x64, Linux x86, Windows aarch64, Windows x64, macOS x64"
```

rkennke:
> I think it would be enough to just trim the defaults here:
> 
> ```
>  default: "Linux additional (hotspot only), Linux x64, Linux x86, Windows aarch64, Windows x64, macOS x64"
> ```

That line only exists in the removed submit.yml. I am trying to trim the 3 unsupported arches in build-cross-compile.yml without removing the actual configurations. Let's see if that works. Also, let me know if you think there is a better way. I would like to keep the linux-aarch64 x-compile target.

shipilev (approved):

PR #66, opened 2023-01-23
8300527: [Lilliput] Simplify and optimize loading of Klass*

rkennke:
With the new fast-locking, we can greatly simplify the loading of the Klass* from an object. We no longer have to call into the runtime to ensure that the header is changed under our feet (by installing the INFLATION marker). Instead, we can do a simple check for the object being monitor-locked, and safely load the displaced header, if that is the case.

---------
### Progress

### Issue
 * [JDK-8300527](https://bugs.openjdk.org/browse/JDK-8300527): [Lilliput] Simplify and optimize loading of Klass*

### Reviewers
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - Committer) ⚠️ Review applies to [83c65c16](https://git.openjdk.org/lilliput/pull/66/files/83c65c16f3abaaa1995fa6cf44b103b751cc405d)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput pull/66/head:pull/66` \
`$ git checkout pull/66`

`$ git checkout pull/66` \
`$ git pull https://git.openjdk.org/lilliput pull/66/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 66`

`$ git pr show -t 66`

</details>

<a href="https://git.openjdk.org/lilliput/pull/66.diff">https://git.openjdk.org/lilliput/pull/66.diff</a>

</details>


src/hotspot/share/c1/c1_LIR.cpp
```diff
@@ -888,7 +888,7 @@ void LIR_OpVisitState::visit(LIR_Op* op) {
 
       do_input(opLoadKlass->_obj);
       do_output(opLoadKlass->_result);
-      do_stub(opLoadKlass->_stub);
+      do_stub(opLoadKlass->stub());
```
shipilev, line 891: Minor: why not leave `_stub`?

src/hotspot/share/runtime/synchronizer.cpp
```diff
@@ -795,7 +795,7 @@ struct SharedGlobals {
 
 static SharedGlobals GVars;
 
-markWord ObjectSynchronizer::read_stable_mark(oop obj) {
+static markWord read_stable_mark(oop obj) {
```
shipilev, line 798: Do you still need this method, or? If you still need it, let's not put it outside the `ObjectSynchronizer` "namespace".

shipilev (approved):
Minor nits.

src/hotspot/share/c1/c1_LIR.cpp
rkennke: Right. I changed that back.

src/hotspot/share/runtime/synchronizer.cpp
rkennke: Originally, read_stable_mark() is static in upstream. A while back I put it into the ObjectSynchronizer namespace to reuse it in Shenandoah, but that is no longer needed, so I reverted the scope back to compilation-unit-private. Yes, we will still need it, when we make Lilliput switchable, but I'd keep it compilation-unit-private for the reasons mentioned.

shipilev: All right then!

PR #67, opened 2023-01-24
8300888: [Lilliput] Remaining missing parts in the SA

rkennke:
With fast-locking, it is now possible to provide a safe implementation of load-Klass* (following the same principle as in the runtime: load mark, check for monitor, possibly load real mark from monitor, decode Klass* from mark). This change also fixes array layout calculation and un-blocks a number of SA tests that did not work before.

---------
### Progress

### Issue
 * [JDK-8300888](https://bugs.openjdk.org/browse/JDK-8300888): [Lilliput] Remaining missing parts in the SA

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput pull/67/head:pull/67` \
`$ git checkout pull/67`

`$ git checkout pull/67` \
`$ git pull https://git.openjdk.org/lilliput pull/67/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 67`

`$ git pr show -t 67`

</details>

<a href="https://git.openjdk.org/lilliput/pull/67.diff">https://git.openjdk.org/lilliput/pull/67.diff</a>

</details>

### DIFF
--- a/src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/debugger/DebuggerBase.java
+++ b/src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/debugger/DebuggerBase.java
@@ -24,6 +24,8 @@
 
 package sun.jvm.hotspot.debugger;
 
+import sun.jvm.hotspot.oops.Mark;
+
 /** <P> DebuggerBase is a recommended base class for debugger
     implementations. It can use a PageCache to cache data from the
     target process. Note that this class would not be suitable if the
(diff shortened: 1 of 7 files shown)


src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/debugger/DebuggerBase.java
```diff
@@ -394,7 +396,11 @@ protected long readCompOopAddressValue(long address)
 
   protected long readCompKlassAddressValue(long address)
     throws UnmappedAddressException, UnalignedAddressException {
-    long value = readCInteger(address, getKlassPtrSize(), true);
+    // On 64 bit systems, the compressed Klass* is currently read from the mark
+    // word. We need to load the whole mark, and shift the upper parts.
+    long value = readCInteger(address, getJLongSize(), true);
```
shipilev, line 401: `machDesc.getAddressSize()` might be a better fit here instead of `getJLongSize()`?

shipilev (approved):
Minor nits.

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/debugger/DebuggerBase.java
rkennke: Right. Both is not beautiful, but address size seems less ugly. At some point in the future all of this will go away, anyway.

shipilev: `getAddressSize` does not over-read accidentally on 32-bit :)

shipilev (approved):

PR #73, opened 2023-02-08
8302065: [Lilliput] Cleanups and touch-ups

rkennke:
Lilliput accumulated a couple of superfluous changes and could need some minor improvements.

### DIFF
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.hpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.hpp
@@ -106,7 +106,6 @@ class MacroAssembler: public Assembler {
   static const char* describe_klass_decode_mode(KlassDecodeMode mode);
 
  private:
-
   static KlassDecodeMode _klass_decode_mode;
 
  public:
--- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
@@ -812,7 +812,7 @@ void C2_MacroAssembler::fast_unlock(Register objReg, Register boxReg, Register t
     jccb(Assembler::zero, Stacked);
     if (UseFastLocking) {
       // If the owner is ANONYMOUS, we need to fix it - in the slow-path.
-      testptr(Address(tmpReg, OM_OFFSET_NO_MONITOR_VALUE_TAG(owner)), (int32_t) (intptr_t) ANONYMOUS_OWNER);
+      testb(Address(tmpReg, OM_OFFSET_NO_MONITOR_VALUE_TAG(owner)), (int32_t) (intptr_t) ANONYMOUS_OWNER);
       jcc(Assembler::notEqual, NO_COUNT);
     }
   }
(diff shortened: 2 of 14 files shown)


src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -812,7 +812,7 @@ void C2_MacroAssembler::fast_unlock(Register objReg, Register boxReg, Register t
     jccb(Assembler::zero, Stacked);
     if (UseFastLocking) {
       // If the owner is ANONYMOUS, we need to fix it - in the slow-path.
-      testptr(Address(tmpReg, OM_OFFSET_NO_MONITOR_VALUE_TAG(owner)), (int32_t) (intptr_t) ANONYMOUS_OWNER);
+      testb(Address(tmpReg, OM_OFFSET_NO_MONITOR_VALUE_TAG(owner)), (int32_t) (intptr_t) ANONYMOUS_OWNER);
```
shipilev, line 815: Sanity question: This would accept the monitor that has the lowest byte equal to `ANONYMOUS_OWNER`. Are we guaranteed that no monitor would ever have the lowest bit as `1`? Probably true, just checking the reason here.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.hpp
```diff
@@ -106,7 +106,6 @@ class MacroAssembler: public Assembler {
   static const char* describe_klass_decode_mode(KlassDecodeMode mode);
 
  private:
-
```
shipilev, line 109: Missed a similar cleanup in `macroAssembler_x86.hpp`, I think.

src/hotspot/cpu/x86/vtableStubs_x86_64.cpp
```diff
@@ -217,8 +217,7 @@ VtableStub* VtableStubs::create_itable_stub(int itable_index) {
   // We expect we need index_dependent_slop extra bytes. Reason:
   // The emitted code in lookup_interface_method changes when itable_index exceeds 15.
   // For linux, a very narrow estimate would be 112, but Solaris requires some more space (130).
-  // Lilliput: expanded since load_klass is significantly and decode_klass somewhat larger.
-  const ptrdiff_t estimate = 155;
+  const ptrdiff_t estimate = 137;
```
shipilev, line 220: Upstream has it as `136`. Do we need `137`?

shipilev (approved):
Looks good. I cross-referenced in with https://builds.shipilev.net/patch-openjdk-lilliput/ -- and most are clean reversal. A few questions, though.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
rkennke: Valid monitor owners can only be Thread* or ANONYMOUS or NULL (or stack-pointers, but only with stack-locking). Thread instances are allocated at least word-aligned, so no, 1 cannot be a valid monitor owner.

src/hotspot/cpu/x86/vtableStubs_x86_64.cpp
rkennke: I couldn't build/run it with 136, it asked for 137. *shrugs*

src/hotspot/cpu/aarch64/macroAssembler_aarch64.hpp
rkennke: Right. I removed the newline in that file, too.

shipilev (approved):

PR #74, opened 2023-02-10
8302209: [Lilliput] Optimize fix-anon monitor owner path

rkennke:
When trying to exit a monitor, we need to check if the current owner is anonymous, and if so, fix it before exiting the monitor. So far, we have been doing this by calling into the slow-path and handle it in the runtime. However, it is fairly easy to do in the fast-path and avoid calling into the runtime altogether. I also included those changes in [upstream fast-locking PR](https://github.com/openjdk/jdk/pull/10907).

Testing:
 - [x] tier1 (x86_64, aarch64)
 - [x] tier2 (x86_64, aarch64)
 - [ ] jcstress -t sync -m quick

---------
### Progress

### Issue
 * [JDK-8302209](https://bugs.openjdk.org/browse/JDK-8302209): [Lilliput] Optimize fix-anon monitor owner path

### Reviewers
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - Committer) ⚠️ Review applies to [808ae7e3](https://git.openjdk.org/lilliput/pull/74/files/808ae7e3ba994ef8b105f03c85f73b2feeb08603)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput pull/74/head:pull/74` \
`$ git checkout pull/74`

`$ git checkout pull/74` \
`$ git pull https://git.openjdk.org/lilliput pull/74/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 74`

`$ git pr show -t 74`

</details>

<a href="https://git.openjdk.org/lilliput/pull/74.diff">https://git.openjdk.org/lilliput/pull/74.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/c2_CodeStubs_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/c2_CodeStubs_aarch64.cpp
@@ -75,22 +75,24 @@ void C2CheckLockStackStub::emit(C2_MacroAssembler& masm) {
   __ b(continuation());
 }
 
-int C2FixAnonOMOwnerStub::max_size() const {
+int C2HandleAnonOMOwnerStub::max_size() const {
   return 20;
 }
 
-void C2FixAnonOMOwnerStub::emit(C2_MacroAssembler& masm) {
+void C2HandleAnonOMOwnerStub::emit(C2_MacroAssembler& masm) {
   __ bind(entry());
   Register mon = monitor();
+  Register thr = thread();
   Register t = tmp();
+  assert(t != noreg, "need tmp register");
 
   // Fix owner to be the current thread.
-  __ str(rthread, Address(mon, ObjectMonitor::owner_offset_in_bytes()));
+  __ str(thr, Address(mon, ObjectMonitor::owner_offset_in_bytes()));
 
   // Pop owner object from lock-stack.
-  __ ldr(t, Address(rthread, JavaThread::lock_stack_current_offset()));
+  __ ldr(t, Address(thr, JavaThread::lock_stack_current_offset()));
   __ sub(t, t, oopSize);
-  __ str(t, Address(rthread, JavaThread::lock_stack_current_offset()));
+  __ str(t, Address(thr, JavaThread::lock_stack_current_offset()));
 
   __ b(continuation());
 }
--- a/src/hotspot/cpu/aarch64/c2_CodeStubs_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/c2_CodeStubs_aarch64.cpp
@@ -75,6 +75,26 @@ void C2CheckLockStackStub::emit(C2_MacroAssembler& masm) {
   __ b(continuation());
 }
 
+int C2FixAnonOMOwnerStub::max_size() const {
+  return 20;
+}
+
+void C2FixAnonOMOwnerStub::emit(C2_MacroAssembler& masm) {
+  __ bind(entry());
+  Register mon = monitor();
+  Register t = tmp();
+
+  // Fix owner to be the current thread.
+  __ str(rthread, Address(mon, ObjectMonitor::owner_offset_in_bytes()));
+
+  // Pop owner object from lock-stack.
+  __ ldr(t, Address(rthread, JavaThread::lock_stack_current_offset()));
+  __ sub(t, t, oopSize);
+  __ str(t, Address(rthread, JavaThread::lock_stack_current_offset()));
+
+  __ b(continuation());
+}
+
 int C2LoadNKlassStub::max_size() const {
   return 8;
 }
--- a/src/hotspot/cpu/aarch64/c2_CodeStubs_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/c2_CodeStubs_aarch64.cpp
@@ -82,17 +82,16 @@ int C2HandleAnonOMOwnerStub::max_size() const {
 void C2HandleAnonOMOwnerStub::emit(C2_MacroAssembler& masm) {
   __ bind(entry());
   Register mon = monitor();
-  Register thr = thread();
   Register t = tmp();
   assert(t != noreg, "need tmp register");
 
   // Fix owner to be the current thread.
-  __ str(thr, Address(mon, ObjectMonitor::owner_offset_in_bytes()));
+  __ str(rthread, Address(mon, ObjectMonitor::owner_offset_in_bytes()));
 
   // Pop owner object from lock-stack.
-  __ ldr(t, Address(thr, JavaThread::lock_stack_current_offset()));
+  __ ldr(t, Address(rthread, JavaThread::lock_stack_current_offset()));
   __ sub(t, t, oopSize);
-  __ str(t, Address(thr, JavaThread::lock_stack_current_offset()));
+  __ str(t, Address(rthread, JavaThread::lock_stack_current_offset()));
 
   __ b(continuation());
 }
--- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
@@ -813,10 +813,16 @@ void C2_MacroAssembler::fast_unlock(Register objReg, Register boxReg, Register t
     if (UseFastLocking) {
       // If the owner is ANONYMOUS, we need to fix it.
       testb(Address(tmpReg, OM_OFFSET_NO_MONITOR_VALUE_TAG(owner)), (int32_t) (intptr_t) ANONYMOUS_OWNER);
-      C2HandleAnonOMOwnerStub* stub = new (Compile::current()->comp_arena()) C2HandleAnonOMOwnerStub(tmpReg, thread);
+#ifdef _LP64
+      C2HandleAnonOMOwnerStub* stub = new (Compile::current()->comp_arena()) C2HandleAnonOMOwnerStub(tmpReg);
       Compile::current()->output()->add_stub(stub);
       jcc(Assembler::notEqual, stub->entry());
       bind(stub->continuation());
+#else
+      // We can't easily implement this optimization on 32 bit because we don't have a thread register.
+      // Call the slow-path instead.
+      jcc(Assembler::notEqual, NO_COUNT);
+#endif
     }
   }
 
--- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
@@ -811,9 +811,16 @@ void C2_MacroAssembler::fast_unlock(Register objReg, Register boxReg, Register t
 #endif
     jccb(Assembler::zero, Stacked);
     if (UseFastLocking) {
-      // If the owner is ANONYMOUS, we need to fix it - in the slow-path.
+      // If the owner is ANONYMOUS, we need to fix it.
       testptr(Address(tmpReg, OM_OFFSET_NO_MONITOR_VALUE_TAG(owner)), (int32_t) (intptr_t) ANONYMOUS_OWNER);
+#ifdef _LP64
+      C2FixAnonOMOwnerStub* stub = new (Compile::current()->comp_arena()) C2FixAnonOMOwnerStub(tmpReg);
+      Compile::current()->output()->add_stub(stub);
+      jcc(Assembler::notEqual, stub->entry());
+      bind(stub->continuation());
+#else
       jcc(Assembler::notEqual, NO_COUNT);
+#endif
     }
   }
 
--- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
@@ -813,14 +813,10 @@ void C2_MacroAssembler::fast_unlock(Register objReg, Register boxReg, Register t
     if (UseFastLocking) {
       // If the owner is ANONYMOUS, we need to fix it.
       testb(Address(tmpReg, OM_OFFSET_NO_MONITOR_VALUE_TAG(owner)), (int32_t) (intptr_t) ANONYMOUS_OWNER);
-#ifdef _LP64
-      C2FixAnonOMOwnerStub* stub = new (Compile::current()->comp_arena()) C2FixAnonOMOwnerStub(tmpReg);
+      C2HandleAnonOMOwnerStub* stub = new (Compile::current()->comp_arena()) C2HandleAnonOMOwnerStub(tmpReg, thread);
       Compile::current()->output()->add_stub(stub);
       jcc(Assembler::notEqual, stub->entry());
       bind(stub->continuation());
-#else
-      jcc(Assembler::notEqual, NO_COUNT);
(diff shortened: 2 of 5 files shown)


src/hotspot/share/opto/c2_CodeStubs.hpp
```diff
@@ -95,6 +95,19 @@ class C2CheckLockStackStub : public C2CodeStub {
 };
 
 #ifdef _LP64
+class C2FixAnonOMOwnerStub : public C2CodeStub {
```
tstuefe, line 98: I dislike the name a bit since it does more than fix the owner in monitor, it also pops one slot from the lock stack.

tstuefe (commented):
Questions/Comments:

What is the advantage of using a stub here? The code section between aarch64 and x64 are different and only used in one place. So you don't really save much by reuse here. Seems you could have just written them out as a conditional path. Then you would have saved the jumps into and out of the stub in the case owner==ANONYMOUS case.

Anonymous means some other thread T2 inflated my lock while I (T1) was thin-lock-waiting on the lock, right?

tstuefe:
BTW I found it confusing that we have pre-existing functions called ...fast_unlock and ...fast_lock, since it has nothing to do with your FastLocking, right?

We should rename UseFastLocking to UseRomanLocking :-)

rkennke:
Hi Thomas,

> What is the advantage of using a stub here? The code section between aarch64 and x64 are different and only used in one place each. So you don't really save much by reuse here. Seems you could have just written them out as a conditional path into the various xxfastlock() functions. Then you would have saved the jumps into and out of the stub in the case owner==ANONYMOUS case.

This is not about saving memory. The very common path here is that owner is not ANONYMOUS. Using a stub means that we forward-jump only in the very uncommon path, and not branch in the common path. If we were not using a stub, we would have to forward-jump in the common path and not branch in the uncommon path. In my experience, forward-jumping in the common path throws off static branch prediction and performs significantly worse. This is also mentioned in the Intel optimization guide as important consideration.

> Anonymous means some other thread T2 inflated my lock while I (T1) was thin-lock-waiting on the lock, right?

Yes, exactly.

I'll address your other comments later today or tomorrow.

Thank you!
Roman

tstuefe:
Hi Roman,

> > What is the advantage of using a stub here? The code section between aarch64 and x64 are different and only used in one place each. So you don't really save much by reuse here. Seems you could have just written them out as a conditional path into the various xxfastlock() functions. Then you would have saved the jumps into and out of the stub in the case owner==ANONYMOUS case.
> 
> This is not about saving memory. The very common path here is that owner is not ANONYMOUS. Using a stub means that we forward-jump only in the very uncommon path, and not branch in the common path. 

wait... now I'm confused. Sorry, I'm no compiler guy. But:

```
__ tst(disp_hdr, (uint64_t)(intptr_t) ANONYMOUS_OWNER);
...
 __ br(Assembler::NE, stub->entry());
```
Why NE? we jump into the sub if owner != ANONYMOUS?

> If we were not using a stub, we would have to forward-jump in the common path and not branch in the uncommon path. In my experience, forward-jumping in the common path throws off static branch prediction and performs significantly worse. This is also mentioned in the Intel optimization guide as important consideration.

I understand that. But you could have this done in-place too, right? Place a label at the end of the function and jump to it in the uncommon case, before the label do a ret for the common case. This is more to make sure I understand the coding, if you prefer to do it via stub that's fine.

> 
> > Anonymous means some other thread T2 inflated my lock while I (T1) was thin-lock-waiting on the lock, right?
> 
> Yes, exactly.

Okay, non-anonymous is the case where a thread leaves an inflated monitor it inflated itself. And that is the hot path? Interesting. Under which circumstances does a thread inflate its own thread? Is this because your fastlocking does not work with recursion yet, right? So every recursion would inflate the lock right away?

> 
> I'll address your other comments later today or tomorrow.
> 
> Thank you! Roman

rkennke:
> > > What is the advantage of using a stub here? The code section between aarch64 and x64 are different and only used in one place each. So you don't really save much by reuse here. Seems you could have just written them out as a conditional path into the various xxfastlock() functions. Then you would have saved the jumps into and out of the stub in the case owner==ANONYMOUS case.
> > 
> > 
> > This is not about saving memory. The very common path here is that owner is not ANONYMOUS. Using a stub means that we forward-jump only in the very uncommon path, and not branch in the common path.
> 
> wait... now I'm confused. Sorry, I'm no compiler guy. But:
> 
> ```
> __ tst(disp_hdr, (uint64_t)(intptr_t) ANONYMOUS_OWNER);
> ...
>  __ br(Assembler::NE, stub->entry());
> ```
> Why NE? we jump into the stub if owner != ANONYMOUS?

tst is an and instruction in disguise: it masks the value and sets ZF according to the result: it is ZF=1 if the bit is 0, and ZF=0 if the bit is 1. NE is ZF=0, that is when the ANONYMOUS bit is set.

> > If we were not using a stub, we would have to forward-jump in the common path and not branch in the uncommon path. In my experience, forward-jumping in the common path throws off static branch prediction and performs significantly worse. This is also mentioned in the Intel optimization guide as important consideration.
> 
> I understand that. But you could have this done in-place too, right? Place a label at the end of the function and jump to it in the uncommon case, before the label do a ret for the common case. This is more to make sure I understand the coding, if you prefer to do it via stub that's fine.

There is no ret here. This whole code section is expanded when C2 emits FastLockNode and FastUnlockNode. (To make it even more confusing, FastLockNode and FastUnlockNode are subclasses of CmpNode, and the code following it will do different things depending on whether we come out with ZF=1 or =0. But this is not relevant for this discussion.) Lacking a ret, I don't see a reasonable way to achieve what I described without using a stub.


> > > Anonymous means some other thread T2 inflated my lock while I (T1) was thin-lock-waiting on the lock, right?
> > 
> > 
> > Yes, exactly.
> 
> Okay, non-anonymous is the case where a thread leaves an inflated monitor it inflated itself.

It doesn't have to have inflated that monitor itself, it would only have locked the (possibly pre-existing) monitor itself.

> And that is the hot path? Interesting. Under which circumstances does a thread inflate its own thread? Is this because your fastlocking does not work with recursion yet, right? So every recursion would inflate the lock right away?

Once 2 or more threads compete for a lock, that lock gets inflated. From there on, all locks and unlocks are done on the monitor. Yes, this is very common and hot (in workloads that use locking properly - as opposed to code that uses uncontended and/or unshared locks, which is what fast-locking is good at). Yes, recursion would inflate the lock immediately, but this is not the point here. (I do have an experimental patch to implement recursive fast-locking in https://github.com/rkennke/jdk/pull/2, I am just not sure if it's worth to have this extra complexity for little gain in somewhat bad workloads.)

Roman

tstuefe:
> > 
> > wait... now I'm confused. Sorry, I'm no compiler guy. But:
> > ```
> > __ tst(disp_hdr, (uint64_t)(intptr_t) ANONYMOUS_OWNER);
> > ...
> >  __ br(Assembler::NE, stub->entry());
> > ```
 >   
> > Why NE? we jump into the stub if owner != ANONYMOUS?
> 
> tst is an and instruction in disguise: it masks the value and sets ZF according to the result: it is ZF=1 if the bit is 0, and ZF=0 if the bit is 1. NE is ZF=0, that is when the ANONYMOUS bit is set.

Ah, okay, thanks for clarifying.

> 
> > > If we were not using a stub, we would have to forward-jump in the common path and not branch in the uncommon path. In my experience, forward-jumping in the common path throws off static branch prediction and performs significantly worse. This is also mentioned in the Intel optimization guide as important consideration.
> > 
> > I understand that. But you could have this done in-place too, right? Place a label at the end of the function and jump to it in the uncommon case, before the label do a ret for the common case. This is more to make sure I understand the coding, if you prefer to do it via stub that's fine.
> 
> There is no ret here. This whole code section is expanded when C2 emits FastLockNode and FastUnlockNode. (To make it even more confusing, FastLockNode and FastUnlockNode are subclasses of CmpNode, and the code following it will do different things depending on whether we come out with ZF=1 or =0. But this is not relevant for this discussion.) Lacking a ret, I don't see a reasonable way to achieve what I described without using a stub.

Okay, I get it. There is no "end of function" to push your uncommon branch to, so the only way to do this without a stub would be to create a branch in-place, and that you don't want that for performance reasons. 

> 
> > > > Anonymous means some other thread T2 inflated my lock while I (T1) was thin-lock-waiting on the lock, right?
> > > 
> > > 
> > > Yes, exactly.
> > 
> > 
> > Okay, non-anonymous is the case where a thread leaves an inflated monitor it inflated itself.
> 
> It doesn't have to have inflated that monitor itself, it would only have locked the (possibly pre-existing) monitor itself.
> 
> > And that is the hot path? Interesting. Under which circumstances does a thread inflate its own thread? Is this because your fastlocking does not work with recursion yet, right? So every recursion would inflate the lock right away?
> 
> Once 2 or more threads compete for a lock, that lock gets inflated. From there on, all locks and unlocks are done on the monitor. Yes, this is very common and hot (in workloads that use locking properly - as opposed to code that uses uncontended and/or unshared locks, which is what fast-locking is good at). Yes, recursion would inflate the lock immediately, but this is not the point here. (I do have an experimental patch to implement recursive fast-locking in [rkennke/jdk#2](https://github.com/rkennke/jdk/pull/2), I am just not sure if it's worth to have this extra complexity for little gain in somewhat bad workloads.)

Again, thanks for explaining.
> 
> Roman

tstuefe:
Reading through fast_unlock() some more...

What I did not understand at first is why you pop the object twice in fast_unlock: once in the stub, then again in fast_unlock_impl. But "Stacked" means thin-locked, right? So the section at 948ff is for handling stack-based thin locks?

If that is so, at some point in time we may want to rename things to make it simpler. E.g. "Stacked"->"Thin" or similar. "Thin" would be a good fit for both new-style locks and old stack-based locks. I understand that you don't want to do this downstream though to keep the delta small.

tstuefe:
Could we factor out lock stack pushing and popping into some utility functions?

Also, would it make sense to emit an assert in debug when popping to not get into negative values, similar to what you do in `MacroAssembler::fast_lock_impl`?

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -811,9 +811,16 @@ void C2_MacroAssembler::fast_unlock(Register objReg, Register boxReg, Register t
...
-      // If the owner is ANONYMOUS, we need to fix it - in the slow-path.
+      // If the owner is ANONYMOUS, we need to fix it.
       testb(Address(tmpReg, OM_OFFSET_NO_MONITOR_VALUE_TAG(owner)), (int32_t) (intptr_t) ANONYMOUS_OWNER);
+#ifdef _LP64
+      C2FixAnonOMOwnerStub* stub = new (Compile::current()->comp_arena()) C2FixAnonOMOwnerStub(tmpReg);
+      Compile::current()->output()->add_stub(stub);
+      jcc(Assembler::notEqual, stub->entry());
+      bind(stub->continuation());
+#else
       jcc(Assembler::notEqual, NO_COUNT);
```
tstuefe, line 822: Could you add a comment "// not yet implemented" ? Just to show that there is nothing basic stopping us from doing this for 32-bit, its just not yet done.

src/hotspot/cpu/aarch64/c2_CodeStubs_aarch64.cpp
```diff
@@ -75,6 +75,26 @@ void C2CheckLockStackStub::emit(C2_MacroAssembler& masm) {
...
 }
 
+int C2FixAnonOMOwnerStub::max_size() const {
+  return 20;
+}
+
+void C2FixAnonOMOwnerStub::emit(C2_MacroAssembler& masm) {
+  __ bind(entry());
+  Register mon = monitor();
+  Register t = tmp();
```
tstuefe, line 85: assert that tmp is real, not noreg? Or is there a mechanism that already checks that?

rkennke:
> Stupid question, is it guaranteed that the branch op always works with the location the stub is emitted to? Is there a mechanism to check that its is not too far away? Or is the code heap always small enough for every branch instruction used?

There are asserts in the branch instructions that check that the branch can be encoded. The stubs get emitted right after the method code. Not sure if there is any reasonable way that this could be too far away, but this is the way it is done since forever in C1 and recently also in C2. I would not be too worried about it.
 
> Reading through this and through the justification for biased-locking removal, it seems that which of those two scenarios are more common today is not so clear cut, right? So modern applications see more contented locking and less locking altogether than old apps?

The scenario in which fast-locking (or stack-locking or biased-locking) really helps is actually pretty narrow. Code would have to:
- do uncontended locking
AND
- using different lock objects

Because if it would be using the same lock-object over and over, the cost of allocating ('inflating') a true ObjectMonitor would be quickly amortised. The locking/unlocking primitives with OM should be as fast as with stack-/fast-locking (basically a single CAS plus some small stuff). Only in single-use (or relatively few used) uncontented locks is where fast-/stack-(/biased-)locking would gain anything compared to pure ObjectMonitor locking. Unfortunately, this still happens in a few scenarios: StringBuffer is one such place that is basically guaranteed to be single-threaded, and they're also very often throwaway objects. Collections and I/O streams I would expect to be reused often/long enough to amortise the OM inflation, but some code uses them in a throwaway fashion, too. And then there are cases where synchronized is used in outright stupid nonsense ways.

rkennke:
> Reading through fast_unlock() some more...
> 
> What I did not understand at first is why you pop the object twice in fast_unlock: once in the stub, then again in fast_unlock_impl. But "Stacked" means thin-locked, right? So the section at [c2_MacroAssembler_x86.cpp 948ff](https://github.com/openjdk/lilliput/blob/55b6c241bad84d7aedb7c7a9a874773914ad359c/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp#L948) is for handling stack-based thin locks?

Right. And the other path that handles anonymous owners is dealing with ObjectMonitor based locking.
 
> If that is so, at some point we may want to rename things to make it simpler. E.g. "Stacked"->"Thin" or similar. "Thin" would be a good fit for both new-style locks and old stack-based locks. I understand that you don't want to do this downstream though to keep the delta small.

Yes. That whole code is quite a mess, and I already did some cleanups in upstream, but it still warrants some serious work. In particular, I would like to get rid of that implicit dealing with ZF there.

rkennke:
> Could we factor out lock stack pushing and popping into some utility functions?
> 
> Also, would it make sense to emit an assert in debug when popping to not get into negative values, similar to what you do in `MacroAssembler::fast_lock_impl`?

Yes, but it's creeping scope of this PR. This belongs in the bigger fast-locking PR IMO: https://github.com/openjdk/jdk/pull/10907

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
rkennke: I'll just implement it. *shrugs*

tstuefe (approved):
+1

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
rkennke: In-fact, this is not so easy. See the comment that I added.

PR #76, opened 2023-03-07
8303769: [Lilliput] Fix interpreter asymmetric fast-locking

rkennke:
Currently we get the asymmetric locking check in the interpreters wrong:

ldr(header_reg, Address(rthread, JavaThread::lock_stack_current_offset()));
cmpoop(header_reg, obj_reg);
br(Assembler::NE, slow_case);

The intention is to load the top of the lock-stack, and compare it to the unlocked object, and, if not equal, branch to the slow-path to handle it. However, what it really does is, it loads the *address* of the top of lock-stack, and compares that to the unlocked object. This can never succeed, and therefore we always call the slow-path. Additionally, the address is not the address of the topmost object, it is the address of the next free slot. What we really want to load is the element at -1 oop from that address. This is not incorrect, but it's unnecessarily slow.'

I've already pushed that fix to the upstream fast-locking PR https://github.com/openjdk/jdk/pull/10907

Testing:
 - [x] tier1 (x86_64, x86_32, aarch64)
 - [ ] tier2 (x86_64, x86_32, aarch64)

---------
### Progress

### Issue
 * [JDK-8303769](https://bugs.openjdk.org/browse/JDK-8303769): [Lilliput] Fix interpreter asymmetric fast-locking

### Reviewers
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput pull/76/head:pull/76` \
`$ git checkout pull/76`

`$ git checkout pull/76` \
`$ git pull https://git.openjdk.org/lilliput pull/76/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 76`

`$ git pr show -t 76`

</details>

<a href="https://git.openjdk.org/lilliput/pull/76.diff">https://git.openjdk.org/lilliput/pull/76.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/interp_masm_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/interp_masm_aarch64.cpp
@@ -872,9 +872,10 @@ void InterpreterMacroAssembler::unlock_object(Register lock_reg)
 
       // Check for non-symmetric locking. This is allowed by the spec and the interpreter
       // must handle it.
-      ldr(header_reg, Address(rthread, JavaThread::lock_stack_current_offset()));
-      ldr(header_reg, Address(header_reg, -oopSize));
-      cmpoop(header_reg, obj_reg);
+      Register tmp = header_reg;
+      ldr(tmp, Address(rthread, JavaThread::lock_stack_current_offset()));
+      ldr(tmp, Address(tmp, -oopSize));
+      cmpoop(tmp, obj_reg);
       br(Assembler::NE, slow_case);
 
       ldr(header_reg, Address(obj_reg, oopDesc::mark_offset_in_bytes()));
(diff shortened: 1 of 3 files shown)


tstuefe (commented):
- As with the other PR, reusing named registers for something else confuses casual readers like me :-) consider using aliases within the scopes where they are used. Even if they are just named "tmp".
- Accessing stack -1 looks scary. I would feel better with some sort of emitted assert that we don't access stack -1, unless this would increase the size of compiled code too much or unless we do this test somewhere else.
- You also may want to fix riscv

rkennke:
> * As with the other PR, reusing named registers for something else confuses casual readers like me :-) consider using aliases within the scopes where they are used. Even if they are just named "tmp".

Done.

> * Accessing stack -1 looks scary. I would feel better with some sort of emitted assert that we don't access stack -1, unless this would increase the size of compiled code too much or unless we do this test somewhere else.

This raises an interesting question. I suspect that we may need to emit an underflow check there and throw an IllegalMonitorStateException when that happens. I don't think we need to do that in the compilers, because they would reject unbalanced monitorenter/exit bytecodes. But I'd rather tackle this separately, write a testcase for this, etc.

> * You also may want to fix riscv

Done. I hope it's correct, I did the change blindly.

tstuefe (approved):
+1. Thanks for humoring my register name obsession.

rkennke:
Thanks!
/integrate

PR #78, opened 2023-03-08
8303813: [Lilliput] (AArch64) Use tbz instead of tst and br in load_klass()

rkennke:
In the implementation of load_klass(), we have a test-and-branch idiom:

```
tst(dst, markWord::monitor_value);
br(Assembler::EQ, fast);

```
We can make this better:

`tbz(dst, exact_log2(markWord::monitor_value), fast);
`
This is not only smaller and perhaps a little faster, it also has the advantage that it doesn't touch the condition flags. This is relevant in at least one place: in verify_oop() the condition flag is alive across the load_klass() call and one jtreg test (runtime/CheckUnhandledOops/TestVerifyOops.java) is failing because of that.

Testing:
 - [x] tier1

---------
### Progress

### Issue
 * [JDK-8303813](https://bugs.openjdk.org/browse/JDK-8303813): [Lilliput] (AArch64) Use tbz instead of tst and br in load_klass()

### Reviewers
 * [Paul Hohensee](https://openjdk.org/census#phh) (@phohensee - no project role) ⚠️ Review applies to [0e6a576e](https://git.openjdk.org/lilliput/pull/78/files/0e6a576e579f069bbc8995061978f5763a8fa18e)
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - Committer) ⚠️ Review applies to [0e6a576e](https://git.openjdk.org/lilliput/pull/78/files/0e6a576e579f069bbc8995061978f5763a8fa18e)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput pull/78/head:pull/78` \
`$ git checkout pull/78`

`$ git checkout pull/78` \
`$ git pull https://git.openjdk.org/lilliput pull/78/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 78`

`$ git pr show -t 78`

</details>

<a href="https://git.openjdk.org/lilliput/pull/78.diff">https://git.openjdk.org/lilliput/pull/78.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -4254,8 +4254,7 @@ void MacroAssembler::load_method_holder(Register holder, Register method) {
 }
 
 // Loads the obj's Klass* into dst.
-// src and dst must be distinct registers
-// Preserves all registers (incl src, rscratch1 and rscratch2), but clobbers condition flags
+// Preserves all registers (incl src, rscratch1 and rscratch2).
 void MacroAssembler::load_nklass(Register dst, Register src) {
   assert(UseCompressedClassPointers, "expects UseCompressedClassPointers");
 
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -4263,8 +4263,7 @@ void MacroAssembler::load_nklass(Register dst, Register src) {
 
   // Check if we can take the (common) fast path, if obj is unlocked.
   ldr(dst, Address(src, oopDesc::mark_offset_in_bytes()));
-  tst(dst, markWord::monitor_value);
-  br(Assembler::EQ, fast);
+  tbz(dst, exact_log2(markWord::monitor_value), fast);
 
   // Fetch displaced header
   ldr(dst, Address(dst, OM_OFFSET_NO_MONITOR_VALUE_TAG(header)));


src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -4263,8 +4263,7 @@ void MacroAssembler::load_nklass(Register dst, Register src) {
 
   // Check if we can take the (common) fast path, if obj is unlocked.
   ldr(dst, Address(src, oopDesc::mark_offset_in_bytes()));
-  tst(dst, markWord::monitor_value);
-  br(Assembler::EQ, fast);
+  tbz(dst, exact_log2(markWord::monitor_value), fast);
```
phohensee, line 4266: This is a bit fragile because monitor_value must be a power of 2 for it to work.

rkennke: Yes, and it is (and hopefully, always will be - or else that code would need to be changed anyway). exact_log2() has an assert that checks exactly that.

phohensee: Missed the latter. Lgtm.

phohensee (approved):

rkennke:
Thanks!
/integrate

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
tstuefe: one could even make monitor value and friends `const_expr` and use STATIC_ASSERT here. That way, you get a compile error if that assumption fails without paying at runtime.

tstuefe (approved):
Looks good. "fast" is close enough for TBZ. 

You can adapt the comment now; remove the "clobbers condition flags".

Question, you are not concerned here with avoiding branches for the fast path? Is this less of a problem with arm64?

rkennke:
> Looks good. "fast" is close enough for TBZ.
> 
> You can adapt the comment now; remove the "clobbers condition flags".

Will do.

> Question, you are not concerned here with avoiding branches for the fast path? Is this less of a problem with arm64?

I am, very much. I found that some workloads are regressing close to 10% because of forward-branches in load-klass code path. However, the code in MacroAssembler is only used for interpreter and other various code paths, but not for C1 and C2. The latter have their own implementation and there we are using stubs.

PR #79, opened 2023-03-08
8303816: [Lilliput] Use realloc instead of malloc+copy when growing the lock-stack

rkennke:
We currently grow the fast-locking lock-stack by allocating a larger stack and copying over all elements. This can be done simpler and more efficient by using realloc instead of malloc+copy.

I also applied the same change to the upstream fast-locking PR: https://github.com/openjdk/jdk/pull/10907

Testing:
 - [x] tier1

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.org/bylaws#committer))

### Issue
 * [JDK-8303816](https://bugs.openjdk.org/browse/JDK-8303816): [Lilliput] Use realloc instead of malloc+copy when growing the lock-stack

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput pull/79/head:pull/79` \
`$ git checkout pull/79`

Update a local copy of the PR: \
`$ git checkout pull/79` \
`$ git pull https://git.openjdk.org/lilliput pull/79/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 79`

View PR using the GUI difftool: \
`$ git pr show -t 79`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/lilliput/pull/79.diff">https://git.openjdk.org/lilliput/pull/79.diff</a>

</details>

PR #80, opened 2023-03-08
Inline initial LockStack stack

tstuefe:
Inlines the initial lock stack. This gives us:
- Much higher chance of cache hits when accessing lockstack slots via lockstack header.
- postponed malloc call
- we use actually less memory (not that it matters much) per thread since malloc overhead is at least 48 bytes per allocation, probably more.

Also increases size of initial stack to two slots, and uses realloc instead of malloc+free when growing stack.

### DIFF
--- a/src/hotspot/share/runtime/lockStack.hpp
+++ b/src/hotspot/share/runtime/lockStack.hpp
@@ -35,10 +35,11 @@ class OopClosure;
 class LockStack {
   friend class VMStructs;
 private:
-  static const size_t INITIAL_CAPACITY = 1;
+  static const size_t INITIAL_CAPACITY = 2;
   oop* _base;
   oop* _limit;
   oop* _current;
+  oop _initial[INITIAL_CAPACITY];
 
   void grow(size_t min_capacity);
 
--- a/src/hotspot/share/runtime/lockStack.cpp
+++ b/src/hotspot/share/runtime/lockStack.cpp
@@ -31,14 +31,16 @@
 #include "utilities/ostream.hpp"
 
 LockStack::LockStack() :
-  _base(UseFastLocking && !UseHeavyMonitors ? NEW_C_HEAP_ARRAY(oop, INITIAL_CAPACITY, mtSynchronizer) : NULL),
+  _base(UseFastLocking && !UseHeavyMonitors ? _initial : NULL),
   _limit(_base + INITIAL_CAPACITY),
   _current(_base) {
 }
 
 LockStack::~LockStack() {
   if (UseFastLocking && !UseHeavyMonitors) {
-    FREE_C_HEAP_ARRAY(oop, _base);
+    if (_base != _initial) {
+      FREE_C_HEAP_ARRAY(oop, _base);
+    }
   }
 }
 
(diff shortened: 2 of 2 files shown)


rkennke:
Oh this is nice! I've just done #79 but have closed that now in favour of this PR.

I wonder if it would help even more if we move the LockStack structure somewhere near the beginning of the Thread stucture (similar to the GC state, which we also moved close to the top), which would increase the likelyhood of (L1) cache hit even more?

Ideally, the improvement should show up here: https://github.com/rkennke/fastlockbench (maybe add new benchmark(s) to show that it does?)

tstuefe:
> Oh this is nice! I've just done #79 but have closed that now in favour of this PR.
> 
> I wonder if it would help even more if we move the LockStack structure somewhere near the beginning of the Thread stucture (similar to the GC state, which we also moved close to the top), which would increase the likelyhood of (L1) cache hit even more?
> 

I thought so too, but you would have to put this into Thread, not JavaThread. And then you'd have to measure performance, since some other data gets pushed into the next cache line. And also, of course, there's false sharing to consider. Lockstack is only used from the current thread, right? So placing it together with other such data may be fine.

Note that using three pointers for this structure is a bit wasteful. You could save one 64-bit slot using ints for stack top and stack limit. But I did not feel like rewriting the using code :-)

> Ideally, the improvement should show up here: https://github.com/rkennke/fastlockbench (maybe add new benchmark(s) to show that it does?)

Yes, probably. I did not run benchmarks though.

src/hotspot/share/runtime/lockStack.hpp
```diff
@@ -35,10 +35,11 @@ class OopClosure;
 class LockStack {
   friend class VMStructs;
 private:
-  static const size_t INITIAL_CAPACITY = 1;
```
rkennke, line 38: Uh where did the '1' here come from? I certainly wanted to make this value at least 4. 4 seems to capture a majority of locking depths while not being excessively large. With just 1 or 2 I would expect to blow out the caching advantage with a significant number of workloads. This is probably a left-over from when I wanted to test the growing parts, which I haven't reverted. Please be generous and make that initial size 3 or 4, unless you have objections to that.

rkennke (changes requested):
Patch looks good, but see one comment below, which addresses my own fault. ;-)

Also, I did a quick run with https://github.com/rkennke/fastlockbench and have not seen any regressions (I wouldn't have expected any) but also no significant wins. However, the benchmark is still very simple and doesn't present a workload that pressures caches at all. I would expect that it quickly pulls everything into L1 and go fast from there.

src/hotspot/share/runtime/lockStack.hpp
tstuefe: Done

rkennke (approved):
Looks good, thanks!

rkennke:
FWIW, I've applied the change to upstream fast-locking PR: https://github.com/openjdk/jdk/pull/10907/commits/3beabad3599291f02425be17a5168effb5552e94

tstuefe:
thanks @rkennke 

/integrate

rkennke:
Can you do me a favour and port this to https://github.com/openjdk/lilliput-jdk17u please?
Thanks, Roman

PR #86, opened 2023-04-26
JDK-8307002: [Lilliput] [Metaspace] Make use of Klass alignment gaps

tstuefe:
https://github.com/openjdk/lilliput/pull/13 changed nKlass encoding by increasing the Klass alignment gap to 512. Works fine and gives us a nKlass-value-point-to-class ratio of about 1.5:1, which means we get close to the goal of representing each class with a single value of nKlass without changing Klass to a fixed-sized Layout. It has two disadvantages, though. The most obvious one - the memory waste caused by alignment shadows - is addressed by this issue. 

(The second disadvantage that I don't address in this issue is what John Rose called "naive overalignment", but it has effects on the patch, see below)

That waste matters little for normal programs: the average waste is 256 bytes per class, and if Lilliput saves 4 bytes per object, we break even if we create more than 64 objects per class on average. We usually do that.

But class-generation scenarios are a different matter. There, we often only have a single or very few objects per class, and many many more classes, so the numbers don't add up anymore. So we need a way to eliminate the gaps.

The simple solution is to save non-class metadata from the same class loader into these gaps. This works fine since much of our non-class metadata is very fine granular. Like sand into cracks, they settle into the gaps and fill them nicely. And it is not even complicated since we already have the machinery (the descendant of Coleen's original dark-matter-prevention-structure that we use to manage prematurely deallocated metaspace). 

----

Numbers:

Memory usage is almost back to baseline now. Volkers insane lambda generation test from JDK-8302154, loading 1 million lambdas, -Xshare:off, shows a reduction of 1.74GB->1.32GB, very close to the original 1.31GB baseline:

(interesting because scenarios like these will be the worst case from a metaspace perspective - metaspace usage of normal apps is unexciting with or without Lilliput):

```
Lilliput unpatched:
Virtual space:                                    
  Non-class space:      832,00 MB reserved,     807,12 MB ( 97%) committed,  13 nodes.
      Class space:        1,00 GB reserved,     977,31 MB ( 95%) committed,  1 nodes.                                                     
             Both:        1,81 GB reserved,       1,74 GB ( 96%) committed. 

Lilliput patched:
Virtual space:                                    
  Non-class space:      384,00 MB reserved,     370,75 MB ( 97%) committed,  6 nodes.
      Class space:        1,00 GB reserved,     977,31 MB ( 95%) committed,  1 nodes.                                                     
             Both:        1,38 GB reserved,       1,32 GB ( 96%) committed. 

Stock VM:
Virtual space:                                    
  Non-class space:      832,00 MB reserved,     822,44 MB ( 99%) committed,  13 nodes.
      Class space:        1,00 GB reserved,     521,31 MB ( 51%) committed,  1 nodes.
             Both:        1,81 GB reserved,       1,31 GB ( 72%) committed. 
```

----

Patch:

https://github.com/openjdk/lilliput/pull/13 introduced the notion that every MetaspaceArena has an alignment - instead, as in the stock VM - working with one global alignment. This patch reverts much of this code (makes it easier to merge too). Instead, it treats alignment as an allocation option. An arena can satisfy allocations with different alignment requirements and will do its best to salvage the alignment waste.

That is the core of the patch. The rest is fluff and tests.

---

Notes:

Numbers are somewhat fluid: at the moment, we use a Klass alignment of 512 and a nKlass size of 21 bits. We will probably change nKlass width to 24 or 26 bits for Lilliputs first milestone.

And once this patch is in, the disadvantage of memory waste is eliminated, so we could increase the Klass alignment. Both 512 and 1K are good values: on average a Klass is between 400b...1K. A 1K alignment would spread out the nKlass value points more. Shift the nKlass-value-point-to-class ratio closer to 1:1.

The still unsolved problem is naive overalignment. We may change the encoding scheme of Klass*->nKlass to take that into account. But not in this patch. But this uncertainty is the reason why I opted for a new method `MetaspaceArena::allocate_for_klass` instead of just giving the existing `Metaspace::allocate` method an alignment parameter.

---------
### Progress

### Issue
 * [JDK-8307002](https://bugs.openjdk.org/browse/JDK-8307002): [Lilliput] [Metaspace] Make use of Klass alignment gaps

### Reviewers
 * [Roman Kennke](https://openjdk.org/census#rkennke) (@rkennke - **Reviewer**) ⚠️ Review applies to [3338fb0e](https://git.openjdk.org/lilliput/pull/86/files/3338fb0e71d27c8fd5586f86225945f8b9a36d6f)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/86/head:pull/86` \
`$ git checkout pull/86`

`$ git checkout pull/86` \
`$ git pull https://git.openjdk.org/lilliput.git pull/86/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 86`

`$ git pr show -t 86`

</details>

<a href="https://git.openjdk.org/lilliput/pull/86.diff">https://git.openjdk.org/lilliput/pull/86.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/86#issuecomment-1525925791)

### DIFF
--- a/src/hotspot/share/memory/metaspace/binList.hpp
+++ b/src/hotspot/share/memory/metaspace/binList.hpp
@@ -32,7 +32,6 @@
 #include "utilities/debug.hpp"
 #include "utilities/globalDefinitions.hpp"
 
-
 namespace metaspace {
 
 // BinList is a data structure to manage small to very small memory blocks
--- a/src/hotspot/share/memory/metaspace/binList.hpp
+++ b/src/hotspot/share/memory/metaspace/binList.hpp
@@ -32,6 +32,7 @@
 #include "utilities/debug.hpp"
 #include "utilities/globalDefinitions.hpp"
 
+
 namespace metaspace {
 
 // BinList is a data structure to manage small to very small memory blocks
--- a/src/hotspot/share/memory/metaspace/freeBlocks.cpp
+++ b/src/hotspot/share/memory/metaspace/freeBlocks.cpp
@@ -27,6 +27,7 @@
 #include "memory/metaspace/freeBlocks.hpp"
 #include "utilities/debug.hpp"
 #include "utilities/globalDefinitions.hpp"
+#include "utilities/ostream.hpp"
 
 namespace metaspace {
 
@@ -69,6 +70,11 @@ MetaWord* FreeBlocks::remove_block(size_t requested_word_size) {
     }
   }
   return p;
+}
+
+void FreeBlocks::print_on(outputStream* st) const {
+  st->print("small_blocks: %u blocks, " SIZE_FORMAT " words; tree: %u nodes, " SIZE_FORMAT " words.",
+            _small_blocks.count(), _small_blocks.total_size(), _tree.count(), _tree.total_size());
 }
 
 } // namespace metaspace
--- a/src/hotspot/share/memory/metaspace/freeBlocks.cpp
+++ b/src/hotspot/share/memory/metaspace/freeBlocks.cpp
@@ -31,7 +31,8 @@
 namespace metaspace {
 
 void FreeBlocks::add_block(MetaWord* p, size_t word_size) {
-  assert(word_size >= MinWordSize, "sanity (" SIZE_FORMAT ")", word_size);
+  assert(is_aligned(p, BytesPerWord), "Bad alignment");
+  assert(word_size >= MinWordSize, "invalid block size (" SIZE_FORMAT ")", word_size);
   if (word_size > MaxSmallBlocksWordSize) {
     _tree.add_block(p, word_size);
   } else {
@@ -47,7 +48,12 @@ MetaWord* FreeBlocks::remove_block(size_t requested_word_size) {
   if (requested_word_size > MaxSmallBlocksWordSize) {
     p = _tree.remove_block(requested_word_size, &real_size);
   } else {
+    // Small blocks: prefer bin lists; if nothing fitting found, take a bigger block
+    // and reuse that.
     p = _small_blocks.remove_block(requested_word_size, &real_size);
+    if (p == nullptr) {
+      p = _tree.remove_block(requested_word_size, &real_size);
+    }
   }
   if (p != nullptr) {
     // Blocks which are larger than a certain threshold are split and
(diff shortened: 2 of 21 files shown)


src/hotspot/share/memory/metaspace/freeBlocks.cpp
```diff
@@ -47,7 +48,12 @@ MetaWord* FreeBlocks::remove_block(size_t requested_word_size) {
   if (requested_word_size > MaxSmallBlocksWordSize) {
     p = _tree.remove_block(requested_word_size, &real_size);
   } else {
+    // Small blocks: prefer bin lists; if nothing fitting found, take a bigger block
+    // and reuse that.
     p = _small_blocks.remove_block(requested_word_size, &real_size);
+    if (p == nullptr) {
+      p = _tree.remove_block(requested_word_size, &real_size);
+    }
```
tstuefe, line 56: Reviewer info: this is a functional change, arguably a bug upstream: 
`FreeBlocks` manages small blocks in a linear bin list (`_small_blacks`), larger blocks in a binary tree (`_tree`). Before, when trying to allocate a small block from `FreeBlocks`, we only would look into the bin list. But it makes sense to look into the large block tree too if that fails. Later down, we chop the block in two and save off the remainder anyway.

src/hotspot/share/memory/metaspace/metaspaceAlignment.hpp
```diff
@@ -45,24 +45,20 @@ static const int MetaspaceMinAlignmentWords = MetaspaceMinAlignmentBytes / Bytes
...
-  // a minimal size
-  size_t raw_word_size = MAX2(net_word_size, FreeBlocks::MinWordSize);
+  // Metaspace allocations are aligned to the minimum metaspace alignment
+  // (1 word on 64-bit, 2 words on 32-bit)
+  word_size = align_up(word_size, metaspace::MetaspaceMinAlignmentWords);
 
-  raw_word_size = align_up(raw_word_size, alignment_words);
-
-  return raw_word_size;
+  return word_size;
```
tstuefe, line 61: Reviewer note: both versions of `get_raw_word_size_for_requested_word_size` are equivalent save for the fact that the old one took alignment as a second input parameter.

src/hotspot/share/memory/metaspace/binList.hpp
```diff
@@ -32,6 +32,7 @@
 #include "utilities/debug.hpp"
 #include "utilities/globalDefinitions.hpp"
 
+
```
rkennke, line 35: stray newline?

rkennke (approved):
I like it! I also like that you now preserve legacy behaviour when running with -UseCompactObjectHeaders :-) Change looks good, just a small nit.

tstuefe:
> I like it! I also like that you now preserve legacy behaviour when running with -UseCompactObjectHeaders :-) Change looks good, just a small nit.

Thank you Roman!

PR #88, opened 2023-05-03
8307392: [Lilliput] Revert deflation of dead object's monitors

rkennke:
In https://github.com/openjdk/lilliput/pull/28, I introduced some code that forces deflation of monitors on objects that are about to die. Discussions around a possible upstreaming of this code (https://github.com/openjdk/jdk/pull/13721) comes to the conclusion that this should not be necessary, and indeed, I could not reproduce the original problem that I was having in Lilliput with the code removed. I assume some changes in upstream and/or Lilliput since then have fixed the root cause. Let's remove that code and see what explodes.

Testing: (All with -XX:+UseHeavyMonitors to amplify monitor traffic)
 - [x] hotspot_gc
 - [x] tier1
 - [x] tier2
 - [x] SPECjvm workloads

---------
### Progress

### Issue
 * [JDK-8307392](https://bugs.openjdk.org/browse/JDK-8307392): [Lilliput] Revert deflation of dead object's monitors

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/88/head:pull/88` \
`$ git checkout pull/88`

`$ git checkout pull/88` \
`$ git pull https://git.openjdk.org/lilliput.git pull/88/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 88`

`$ git pr show -t 88`

</details>

<a href="https://git.openjdk.org/lilliput/pull/88.diff">https://git.openjdk.org/lilliput/pull/88.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/88#issuecomment-1533619413)

### DIFF
--- a/src/hotspot/share/gc/shared/oopStorage.inline.hpp
+++ b/src/hotspot/share/gc/shared/oopStorage.inline.hpp
@@ -29,7 +29,6 @@
 
 #include "memory/allocation.hpp"
 #include "oops/oop.hpp"
-#include "runtime/objectMonitor.hpp"
 #include "runtime/safepoint.hpp"
 #include "utilities/align.hpp"
 #include "utilities/count_trailing_zeros.hpp"
@@ -263,7 +262,6 @@ public:
       if (_is_alive->do_object_b(v)) {
         result = _f(ptr);
       } else {
-        ObjectMonitor::maybe_deflate_dead(ptr);
         *ptr = nullptr;            // Clear dead value.
       }
     }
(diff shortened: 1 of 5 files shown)


shipilev (commented):
But we know this would not be safe without the GC threads rendezvous-ing, right? Also, this seems to include #90.

rkennke:
> But we know this would not be safe without the GC threads rendezvous-ing, right? Also, this seems to include #90.

Yes, but in Lilliput we rendezous GC threads since #27. This PR essentially reverts that earlier PR. I removed the Shenandoah part.

shipilev (approved):
Okay then!

PR #92, opened 2023-05-09
JDK-8307737: [Lilliput] Reduce number of loads used for Klass decoding in static code

tstuefe:
Klass decode depends on several runtime parameters. In 64-bit header mode, these are:

- UseCompactObjectHeaders
- Encoding Base
- Encoding Shift

In Legacy header mode, these are:

- UseCompactObjectHeaders
- UseCompressedClassPointers
- Encoding Base
- Encoding Shift

These values are stored at distinct locations and require three resp. four loads (see disassmbly [1]). Unfortunately, Legacy mode is more expensive with Lilliput, since we now need to load two switches.

I want to minimize the number of loads. There are several ways to do this, but the most simple would be to use a denser representation in memory of these values. We always load them together anyway.

All four values (UseCompactObjectHeaders, UseCompressedClassPointers, Encoding Base+Shift) can be coded into a single 64-bit value. The encoding base will always be page-aligned. That leaves us an alignment shadow of 12 bits to hide all the rest of the information. UseCompactObjectHeaders and UseCompressedClassPointers can be represented by single bits. The encoding shift will not be larger than 31, so we can store the shift in 5 bits.

The result is that the three resp. four loads can be folded into a single 64-bit load without too much trouble.

Alternatives. 

- Generating a stub routine is not an option if one wants to keep decoding inlined
- We could generate different variants of the decoding routines via template, parametrized for each permutation of (shift, UseCompressedClassPointers, UseCompactObjectHeaders) and the most common encoding base. But: 
	- we would need a different solution for uncommon base addresses
	- we may want to make shift an adjustable runtime parameter
	- we would need to decide, at runtime, which code variant to use - again, introduces a runtime switch we need to load from memory - nothing gained compared to the proposed solution.

----------------

The patch changes the way the static helper class `CompressedKlassPointers` stores encoding base and shift to a denser format. The format also contains copies of UseCompressedClassPointers and UseCompactObjectHeaders:

```
Bit#
0:	UseCompactObjectHeaders
1:	UseCompressedClassPointers
2-6:	Encoding shift (5 bits)
7-11:   Unused
12-63:  Encoding Base address
```

Patch then changes some frequently used oop methods to use the dense representation of UseCompactObjectHeaders/UseCompressedClassPointers.

The result: we now only need one load for Klass decoding/encoding (see disassembly [2]).

-----------------------

[1] https://bugs.openjdk.org/secure/attachment/103772/KlassExtraction.txt
[2] https://bugs.openjdk.org/secure/attachment/103773/KlassExtraction-patched.txt

---------
### Progress

### Issue
 * [JDK-8307737](https://bugs.openjdk.org/browse/JDK-8307737): [Lilliput] Reduce number of loads used for Klass decoding in static code

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/92/head:pull/92` \
`$ git checkout pull/92`

`$ git checkout pull/92` \
`$ git pull https://git.openjdk.org/lilliput.git pull/92/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 92`

`$ git pr show -t 92`

</details>

<a href="https://git.openjdk.org/lilliput/pull/92.diff">https://git.openjdk.org/lilliput/pull/92.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/92#issuecomment-1540173467)

### DIFF
--- a/src/hotspot/share/oops/compressedKlass.cpp
+++ b/src/hotspot/share/oops/compressedKlass.cpp
@@ -97,6 +97,7 @@ void CompressedKlassPointers::initialize(address addr, size_t len) {
 
   assert(LogKlassAlignmentInBytes < (1 << encodingShiftWidth), "Shift too large");
   assert((((uintptr_t)thebase) & ~baseAddressMask) == 0, "Base address " PTR_FORMAT " unaligned", p2i(thebase));
+
   _value = (UseCompactObjectHeaders ? ((uintptr_t)1 << useCompactObjectHeadersShift) : 0) |
            (UseCompressedClassPointers ? ((uintptr_t)1 << useCompressedClassPointersShift) : 0) |
            ((uintptr_t)LogKlassAlignmentInBytes << encodingShiftShift) |
--- a/src/hotspot/share/oops/compressedKlass.cpp
+++ b/src/hotspot/share/oops/compressedKlass.cpp
@@ -31,7 +31,7 @@
 #include "utilities/debug.hpp"
 #include "runtime/globals.hpp"
 
-uintptr_t CompressedKlassPointers::_value = 0;
+uintptr_t CompressedKlassPointers::_config = 0;
 address CompressedKlassPointers::_base_copy = nullptr;
 int CompressedKlassPointers::_shift_copy = 0;
 
@@ -98,7 +98,7 @@ void CompressedKlassPointers::initialize(address addr, size_t len) {
   assert(LogKlassAlignmentInBytes < (1 << encodingShiftWidth), "Shift too large");
   assert((((uintptr_t)thebase) & ~baseAddressMask) == 0, "Base address " PTR_FORMAT " unaligned", p2i(thebase));
 
-  _value = (UseCompactObjectHeaders ? ((uintptr_t)1 << useCompactObjectHeadersShift) : 0) |
+  _config = (UseCompactObjectHeaders ? ((uintptr_t)1 << useCompactObjectHeadersShift) : 0) |
            (UseCompressedClassPointers ? ((uintptr_t)1 << useCompressedClassPointersShift) : 0) |
            ((uintptr_t)LogKlassAlignmentInBytes << encodingShiftShift) |
            ((uintptr_t)thebase & baseAddressMask);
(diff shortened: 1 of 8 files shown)


src/hotspot/share/oops/compressedKlass.cpp
```diff
@@ -71,25 +75,39 @@ void CompressedKlassPointers::initialize(address addr, size_t len) {
...
+                              p2i(thebase), p2i(addr), p2i(addr + len));
 
   // For SA
+  _base_copy = thebase;
   _shift_copy = LogKlassAlignmentInBytes;
 
+  assert(LogKlassAlignmentInBytes < (1 << encodingShiftWidth), "Shift too large");
+  assert((((uintptr_t)thebase) & ~baseAddressMask) == 0, "Base address " PTR_FORMAT " unaligned", p2i(thebase));
+
+  _value = (UseCompactObjectHeaders ? ((uintptr_t)1 << useCompactObjectHeadersShift) : 0) |
```
rkennke, line 101: I am not sure that _value is a good name. Maybe _config or so?

src/hotspot/share/oops/compressedKlass.hpp
```diff
@@ -66,11 +66,29 @@ class CompressedKlassPointers : public AllStatic {
...
+  static constexpr int encodingShiftShift = 2;
+  static constexpr int encodingShiftWidth = 5;
+  static constexpr int baseAddressMask = ~right_n_bits(12);
+
+  static void set_encoding_base(address base);
+  static void set_encoding_shift(int shift);
+  static void set_use_compact_headers(bool b);
+  static void set_use_compressed_class_pointers(bool b);
+
+  // These members hold copies of encoding base and shift and only exist for SA (see vmStructs.cpp and
```
rkennke, line 89: I guess, SA could be changed to decode the config bits? Maybe not in this PR, though.

rkennke (approved):
Looks good. Some relatively minor comments, up to you if you want to change anything.

src/hotspot/share/oops/compressedKlass.hpp
tstuefe: Yes, we should fix that.

src/hotspot/share/oops/compressedKlass.cpp
tstuefe: Okay, _config it is.

rkennke (approved):
Thanks, LGTM!

PR #93, opened 2023-05-12
8307991: [Lilliput] Use narrow i-hash only when +UseCompactObjectHeaders

rkennke:
Currently, we use a narrow 25bit identity-hash-code in Lilliput. That is even the case when running without compact object headers, even though we could just as well use the original 31bit identity hash-code then.

The fix is relatively simple. Also, this brings the code in line with what we're currently proposing in upstream https://github.com/openjdk/jdk/pull/13844.

Yes, going forward we would trade Klass* bits to get the full 31 i-hash back, and we're going to do that soon, but let's first fix this issue at hand (also in preparation for a clean backport, where we don't currently have the metaspace improvements).

Testing:
 - [ ] tier1 +UseCompactObjectHeaders
 - [ ] tier2 +UseCompactObjectHeaders
 - [ ] tier1 -UseCompactObjectHeaders
 - [ ] tier2 -UseCompactObjectHeaders


shipilev (approved):
Looks good.

PR #102, opened 2023-08-10
8314130: [Lilliput] Make loadNKlassCompactHeaders not use a TEMP registers

rkennke:
The loadNKlassCompactHeaders template in the .ad files currently declares dst as TEMP_DEF. This has historic reasons: We needed to ensure that dst doesn't overlap with src, because we needed src for calling into the slow-path. In addition to that, there seems to be a related bug in C2's matcher which sometimes seems to create an oopmap entry for a temp reg, which would make GCs crash if they scan uninitialized slot (see [JDK-8051805](https://bugs.openjdk.org/browse/JDK-8051805)). That problem should also go away with this change. In addition to that, relaxing the type of dst might also lead to performance enhancements when register pressure is high, because we don't need two registers for loadNKlass - 1 register would be enough.

Testing:
- [x] tier1 (x86_64, aarch64)
- [x] tier2 (x86_64, aarch64)


shipilev (approved):
Okay, this makes sense. 

GHA are broken, because we have not picked up stuff from upstream recently.

PR #111, opened 2023-10-10
8317812: [Lilliput] Make C2 LoadNKlassCompactHeader more robust

rkennke:
Lilliput's C2 code for generating LoadNKlass currently assumes that the disp of the incoming address is klass_offset_in_bytes. It then extracts the base register and loads from the mark_offset_in_bytes instead.
Sometimes (apparently very rarely) it happens that C2 emits code that pre-adds obj+klass_offset into a register, and uses that as base, but with offset 0. In this case we would trip the assert or crash in release builds.

### DIFF
--- a/src/hotspot/cpu/aarch64/aarch64.ad
+++ b/src/hotspot/cpu/aarch64/aarch64.ad
@@ -7346,9 +7346,7 @@ instruct loadNKlassCompactHeaders(iRegNNoSp dst, memory4 mem, rFlagsReg cr)
   ins_cost(4 * INSN_COST);
   format %{ "ldrw  $dst, $mem\t# compressed class ptr" %}
   ins_encode %{
-    assert($mem$$disp == oopDesc::klass_offset_in_bytes(), "expect correct offset");
-    assert($mem$$index$$Register == noreg, "expect no index");
-    __ load_nklass_compact($dst$$Register, $mem$$base$$Register);
+    __ load_nklass_compact($dst$$Register, $mem$$base$$Register, $mem$$index$$Register, $mem$$scale, $mem$$disp);
   %}
   ins_pipe(pipe_slow);
 %}
--- a/src/hotspot/cpu/aarch64/c2_MacroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/c2_MacroAssembler_aarch64.cpp
@@ -2059,10 +2059,25 @@ bool C2_MacroAssembler::in_scratch_emit_size() {
   return MacroAssembler::in_scratch_emit_size();
 }
 
-void C2_MacroAssembler::load_nklass_compact(Register dst, Register obj) {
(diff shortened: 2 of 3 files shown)


src/hotspot/cpu/x86/x86_64.ad
```diff
@@ -5333,9 +5333,9 @@ instruct loadNKlassCompactHeaders(rRegN dst, indOffset8 mem, rFlagsReg cr)
   ins_cost(125); // XXX
   format %{ "movl    $dst, $mem\t# compressed klass ptr" %}
   ins_encode %{
-    assert($mem$$disp == TypeOopPtr::klass_offset_in_bytes(), "expect correct offset 4, but got: %d", $mem$$disp);
-    assert($mem$$index == 4, "expect no index register: %d", $mem$$index);
-    __ load_nklass_compact_c2($dst$$Register, $mem$$base$$Register);
+    Register index = $mem$$index != 4 ? $mem$$index$$Register : noreg;
```
shipilev, line 5336: I don't understand what`$mem$$index != 4` is supposed to check. Can't we just pass the whole register to `load_nklass_compact_c2`? Is it already `noreg` when index is not set?

shipilev: Or is unset index `-1`, not `4`?

rkennke: All the operand definitions use 0x4 for 'no index' e.g.:
operand stackSlotI(sRegI reg)
%{
  constraint(ALLOC_IN_RC(stack_slots));
  // No match rule because this operand is only generated in matching

  format %{ "[$reg]" %}
  interface(MEMORY_INTER) %{
    base(0x4);   // RSP
    index(0x4);  // No Index
    scale(0x0);  // No Scale
    disp($reg);  // Stack Offset
  %}
%}

There are many more cases like this in x86_64

shipilev: On the other hand, this suggests that we can pass `$mem$$index$Register` right away?

https://github.com/openjdk/lilliput/pull/111/files#diff-018aa61d1a7aafcf70a535fcd40a318a4bd6511fd40ac39ce4be90cc52216749L7350

shipilev: In fact, I don't quite see why `x86_64.ad` cannot do what `aarch64.ad` does:

```
__ load_nklass_compact($dst$$Register, $mem$$base$$Register, $mem$$index$$Register, $mem$$scale, $mem$$disp);
```

rkennke: Problem is the index=4 does not decode to noreg (it decodes to rsp) and scale=0 does not decode to no_scale (it's invalid, I think). Hence the extra hoops there.

shipilev: Ooof. Okay then.

shipilev (approved):

shipilev (approved):

PR #122, opened 2024-01-08
[Lilliput] Implement new Serial Full GC

rkennke:
This implements a compressor-style compacting full-GC for the Serial GC, which does not require storing any forwarding pointers in object headers.


rkennke:
The proposed new full GC wouldn't work with compact identity hashcode, and is therefore not useful.

PR #142, opened 2024-03-04
Merge jdk:jdk-23+12

rkennke:
Let's merge jdk-23+12, in particular, this brings in: [8139457: Relax alignment of array elements](https://bugs.openjdk.org/browse/JDK-8139457). After this merge, I'll check the remaining diff vs upstream and see if I find any leftovers.

Testing:
 - [x] tier1 (-UCOH)
 - [x] tier1 (+UCOH)


rkennke:
I see 3 failures with +UCOH, but all of them are preexisting. No failures with -UCOH. Going to push the merge.

PR #149, opened 2024-04-19
Parallel: modify to-space object in array chunking

stefank:
The ParallelGC chunks arrays so that it can somewhat parallelize the array processing. This is done by rewriting the length field of the from-space object. This causes a problem for Lilliput when one thread copies an object, forwards it, and then rewrites the length of the from-space copy, while another thread tries to figure out if it should copy the object and tries to read the size of the object.

The second thread tries to stay away from calling `oopDesc::klass()` on the from-space object, because the header might have been overwritten by another thread. So, it only reads the mark word, tries to figure out the object has been forwarded, and then calls `size_given_klass` to figure out the size.

The `size_given_klass` has some asserts that checks that the different ways to calculate array sizes are consistent, but since other threads could be updating the length field, there is an escape-hatch that makes that assert less strict. This is the assert:
```
assert(s == klass->oop_size(this) || size_might_change(), "wrong array object size");
```

The problem for Lilliput is that `size_might_change()` rereads the object's klass pointer, which will fail because the header is overwritten with a forwarding pointer.

This can be fixed by passing in the klass or maybe by following the forwarding pointer to find the to-space object and find the klass from there, but I wonder if this all wouldn't be safer if we left the from-space object unmodified and instead changed the length field of the to-space object. In fact, this is exactly what G1 does since many years (which also indicates that the `size_might_change()` really shouldn't be checking for UseG1GC).

The above is a proposal of what this would look like. I'm publishing this here for Lilliput because it fixes a real bug here, but if this is the right way to go forward, then this should likely be upstreamed as well.

I've run this through tier1-3 with both UseParallelGC and UseCompactObjectHeaders enabled, and I'm currently running this through higher Oracle tiers.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/149/head:pull/149` \
`$ git checkout pull/149`

`$ git checkout pull/149` \
`$ git pull https://git.openjdk.org/lilliput.git pull/149/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 149`

`$ git pr show -t 149`

</details>

<a href="https://git.openjdk.org/lilliput/pull/149.diff">https://git.openjdk.org/lilliput/pull/149.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/149#issuecomment-2066659531)

### DIFF
--- a/src/hotspot/share/gc/parallel/psPromotionManager.cpp
+++ b/src/hotspot/share/gc/parallel/psPromotionManager.cpp
@@ -280,34 +280,34 @@ template <class T> void PSPromotionManager::process_array_chunk_work(
 void PSPromotionManager::process_array_chunk(PartialArrayScanTask task) {
   assert(PSChunkLargeArrays, "invariant");
 
-  oop old = task.to_source_array();
-  assert(old->forward_safe_klass()->is_objArray_klass(), "invariant");
-  assert(old->is_forwarded(), "invariant");
+  oop const from_obj = task.to_source_array();
+  assert(from_obj->is_forwarded(), "invariant");
+  assert(from_obj->forwardee()->is_objArray(), "invariant");
+
+  objArrayOop to_obj = objArrayOop(from_obj->forwardee());
 
   TASKQUEUE_STATS_ONLY(++_array_chunks_processed);
 
-  oop const obj = old->forwardee();
-
   int start;
-  int const end = arrayOop(old)->length();
+  int const end = to_obj->length();
   if (end > (int) _min_array_size_for_chunking) {
     // we'll chunk more
     start = end - _array_chunk_size;
     assert(start > 0, "invariant");
-    arrayOop(old)->set_length(start);
-    push_depth(ScannerTask(PartialArrayScanTask(old)));
+    to_obj->set_length(start);
+    push_depth(ScannerTask(PartialArrayScanTask(from_obj)));
     TASKQUEUE_STATS_ONLY(++_array_chunk_pushes);
   } else {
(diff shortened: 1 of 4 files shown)


rkennke (approved):
I think it's reasonable and correct.
I don't feel very comfortable messing with the length-field, though. Going forward, we might consider doing what's done in Shenandoah instead, and encode the chunking in the task object.
https://github.com/openjdk/jdk/blob/177092b952c2135c6f6872c6b64d1e210452d35a/src/hotspot/share/gc/shenandoah/shenandoahTaskqueue.hpp#L66

stefank:
I took one step out the door for a walk and immediately realized that during promotion failure the from-space object becomes a to-space object and we're back to square one.

stefank:
> ... I don't feel very comfortable messing with the length-field, though. Going forward, we might consider doing what's done in Shenandoah instead, and encode the chunking in the task object. https://github.com/openjdk/jdk/blob/177092b952c2135c6f6872c6b64d1e210452d35a/src/hotspot/share/gc/shenandoah/shenandoahTaskqueue.hpp#L66

I agree. FWIW, ZGC also encodes the chunk in the "task":
https://github.com/openjdk/jdk/blob/177092b952c2135c6f6872c6b64d1e210452d35a/src/hotspot/share/gc/z/zMarkStackEntry.hpp#L56

stefank:
I'm closing this PR and suggest that we fix this with #163 instead.

PR #150, opened 2024-04-19
SA: Remove !hasMonitor() assert in getKlass()

stefank:
Small patch to fix SA testing with UCOH.


rkennke (approved):
Oops. My oversight.
Looks good!

stefank:
Thanks for reviewing!
/integrate

PR #125, opened 2024-01-23
8319185: [Lilliput] Enable and fix vectorization tests

rkennke:
The change does two (related) things:

It white-lists the UseCompactObjectHeaders flag in the IR testing framework. Without this, the flag would not be passed through to any IR framework test.
That first change makes 3 tests fail. Those are all vectorization tests, and they are failing because 'small' array types (4-byte-elements and smaller) no longer have their elements aligned at 8-byte-boundaries. This throws off vectorization because this only kicks in on aligned accesses. The mitigation is to disable the failing tests for now. #118 offered a more comprehensive solution, but it was also more intrusive. After some discussion with Emmanuel Peter and Maurizio Cimadamore, we came to the conclusion that disabling the tests for now might be a better approach, especially considering that we may soon no longer need special 4-byte-array-alignment anymore.

Testing:
 - [x] compiler/c2/irTests/TestVectorizationMismatchedAccess.java +UCOH
 - [x] compiler/c2/irTests/TestVectorizationNotRun.java +UCOH
 - [x] compiler/loopopts/superword/TestIndependentPacksWithCyclicDependency.java +UCOH
 - [x] compiler/c2/irTests/TestVectorizationMismatchedAccess.java -UCOH
 - [x] compiler/c2/irTests/TestVectorizationNotRun.java -UCOH
 - [x] compiler/loopopts/superword/TestIndependentPacksWithCyclicDependency.java -UCOH
 - [x] tier1 +UCOH
 - [x] tier1 -UCOH

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.org/bylaws#committer))

### Issue
 * [JDK-8319185](https://bugs.openjdk.org/browse/JDK-8319185): [Lilliput] Enable and fix vectorization tests (**Bug** - P4)

### Reviewers
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - Committer)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/125/head:pull/125` \
`$ git checkout pull/125`

Update a local copy of the PR: \
`$ git checkout pull/125` \
`$ git pull https://git.openjdk.org/lilliput.git pull/125/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
[...]

PR #152, opened 2024-04-22
8319185: [Lilliput] Enable and fix vectorization tests

rkennke:
This re-instates #125 after it has been lost with the Great Reset. It's a clean cherry-pick of that other commit.

### DIFF
--- a/test/hotspot/jtreg/compiler/c2/irTests/TestVectorizationMismatchedAccess.java
+++ b/test/hotspot/jtreg/compiler/c2/irTests/TestVectorizationMismatchedAccess.java
@@ -147,7 +147,8 @@ public class TestVectorizationMismatchedAccess {
     }
 
     @Test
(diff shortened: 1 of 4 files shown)


stefank:
Seems reasonable to me, though I'm not really familiar with this test framework. If this makes the testing cleaner I'd prefer to get this integrated now instead of waiting for a perfect fix/review.

rkennke:
> Seems reasonable to me, though I'm not really familiar with this test framework. If this makes the testing cleaner I'd prefer to get this integrated now instead of waiting for a perfect fix/review.

I had some discussions with Emmanuel Peter and Maurizio Cimadamore somewhere else (perhaps in the array base alignment discussions, but I don't remember, tbh), and this seemed to be the most preferable way forward. None of it is catastrophic or can be done much about, it might result in performance nuisances in some corner cases (e.g. when alignment of src and dst arrays don't match, e.g. when arraycopying from a byte[] to a long[] or such). If/when we do 4-byte-headers, the problem will disappear.

stefank (approved):

PR #159, opened 2024-04-22
OMWorld: ObjectMonitor:enter

xmas92:
This restructures `ObjectMonitor:enter` to fit better with the `LightweightSynchronizer`. Some parts of this could be upstreamed separately. 

This effectively splits up `ObjectMonitor::enter` into three parts. Which LightweightSynchronizer can weave into its `inflate_and_enter`. 

```c++
bool ObjectMonitor::enter(JavaThread* current) {
  // 1. Spinning without blocking out deflation
  if (spin_enter(current)) {
    return true;
  }

  // 2. Try to block deflation
  ObjectMonitorContentionMark contention_mark(this);
  if (enter_is_async_deflating()) {
    return false;
  }

  // 3. Enter while blocking out deflation
  enter_with_contention_mark(current, contention_mark);
  return true;
}
```

It may be worth the effort to document these new methods, and also add better documentation to `inflate_and_enter` which describes the invariants and how `ObjectMonitorContentionMark`, markWord transitions and ObjectMonitor::_owner field (on newly inserted monitors) plays together to guarantee correctness and works. 

This commit is currently based on ffc4c2fd5ba1c28432a5d5e84b8ada95fba7eeb8 without a merge. So GHA is not tested with lilliput. As long as this is mergable without conflicts (and no manual merge is performed) it will also be possible to move this commit down below lilliput on-top of OMWorld directly.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.org/bylaws#committer))

### Reviewers
 * [Roman Kennke](https://openjdk.org/census#rkennke) (@rkennke - **Reviewer**) ⚠️ Review applies to [24ed7d73](https://git.openjdk.org/lilliput/pull/159/files/24ed7d7341881b8eae240fda999021bf5f2159a7)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/159/head:pull/159` \
`$ git checkout pull/159`

Update a local copy of the PR: \
`$ git checkout pull/159` \
`$ git pull https://git.openjdk.org/lilliput.git pull/159/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 159`

View PR using the GUI difftool: \
`$ git pr show -t 159`

</details>
<details><summary>Using diff file</summary>

[...]

PR #160, opened 2024-04-22
OMWorld: Remove BasicLock* restrictions

xmas92:
This is based on #159 

The main idea is to remove the restrictions where the `BasicLock*` may be missing when calling into the runtime. This is only an issue on x86_32. For now simply call into the runtime at the places where the registers are lacking. `quick_enter` should take care of the cases where we otherwise would have skipped calling into the runtime, as x86_32 calls into the runtime regardless (to get the thread) this may even be performance neutral.

---------
### Progress

### Reviewers
 * [Roman Kennke](https://openjdk.org/census#rkennke) (@rkennke - **Reviewer**) ⚠️ Review applies to [5c2981ad](https://git.openjdk.org/lilliput/pull/160/files/5c2981adc5f8bdf87e3c4fd6810f68cc6798f094)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/160/head:pull/160` \
`$ git checkout pull/160`

`$ git checkout pull/160` \
`$ git pull https://git.openjdk.org/lilliput.git pull/160/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 160`

`$ git pr show -t 160`

</details>

<a href="https://git.openjdk.org/lilliput/pull/160.diff">https://git.openjdk.org/lilliput/pull/160.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/160#issuecomment-2072116014)


rkennke (approved):
Nice! Looks reasonable. Yeah I've been thinking the same: if we need to call runtime to fetch the Thread* anyway, we might as well do all the locking stuff there.

src/hotspot/cpu/x86/c1_MacroAssembler_x86.cpp
```diff
@@ -66,10 +67,12 @@ int C1_MacroAssembler::lock_object(Register hdr, Register obj, Register disp_hdr
   if (LockingMode == LM_LIGHTWEIGHT) {
 #ifdef _LP64
     const Register thread = r15_thread;
+    lightweight_lock(disp_hdr, obj, hdr, thread, tmp, slow_case);
 #else
-    const Register thread = disp_hdr;
+    movptr(hdr, Address(obj, oopDesc::mark_offset_in_bytes()));
```
stefank, line 72: Maybe add a comment about this null-check?

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -1208,41 +1208,26 @@ void C2_MacroAssembler::fast_unlock_lightweight(Register obj, Register reg_rax,
...
-        orptr(reg_rax, Address(monitor, ObjectMonitor::cxq_offset()));
-        jcc(Assembler::notZero, check_successor);
+      Label recursive;
 
-        // Release lock.
-        movptr(Address(monitor, ObjectMonitor::owner_offset()), NULL_WORD);
-#else // _LP64
-        Label recursive;
+      // Check if recursive.
+      cmpptr(Address(monitor,ObjectMonitor::recursions_offset()),0);
```
stefank, line 1215:
```suggestion
      cmpptr(Address(monitor,ObjectMonitor::recursions_offset()), 0);
```

coleenp (commented):
This looks applicable to mainline, can it be pushed there?

xmas92:
> This looks applicable to mainline, can it be pushed there?

Yes. 

There are some other changes that could be split out into upstream patches as well. The plan for now is to get this patch queue in, rebase and squash together into one OMWorld commit and then look at what can / should be broken out into upstream patches.

The hope is that after this patch queue the runtime changes are close to final (with exception of the hash table grow and shrinking code).

PR #163, opened 2024-04-23
Parallel: fix size_given_klass

stefank:
The Parallel scavenge threads compete to copy and forward objects, and the winning object is the one that managed to update the the forwarding pointer. For Lilliput, this code needs to be extra carefully written because installing the forwarding pointer destroys the klass pointer. The threads need to read the object header once, and then use that information when figuring out the size of the object. The code has been changed to do this, but there's a place in an assert were the code tries to reread the klass pointer, and then crash because it has been racingly been overwritten by another thread.

I had an idea that we could fix this by simplifying the assert with #149, but that has a problem with self-forwarded objects. There are ways to fix the problem listed in that PR, but for now the suggestion is to just make sure that the asserting code doesn't reread the klass pointer.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/163/head:pull/163` \
`$ git checkout pull/163`

`$ git checkout pull/163` \
`$ git pull https://git.openjdk.org/lilliput.git pull/163/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 163`

`$ git pr show -t 163`

</details>

<a href="https://git.openjdk.org/lilliput/pull/163.diff">https://git.openjdk.org/lilliput/pull/163.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/163#issuecomment-2072091929)

### DIFF
--- a/src/hotspot/share/oops/oop.cpp
+++ b/src/hotspot/share/oops/oop.cpp
@@ -221,12 +221,12 @@ jdouble oopDesc::double_field_acquire(int offset) const               { return A
 void oopDesc::release_double_field_put(int offset, jdouble value)     { Atomic::release_store(field_addr<jdouble>(offset), value); }
 
 #ifdef ASSERT
-bool oopDesc::size_might_change() {
+bool oopDesc::size_might_change(Klass* klass) {
   // UseParallelGC and UseG1GC can change the length field
   // of an "old copy" of an object array in the young gen so it indicates
   // the grey portion of an already copied array. This will cause the first
   // disjunct below to fail if the two comparands are computed across such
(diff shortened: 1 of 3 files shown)


stefank:
I've tested this by running tier1-3 with UCOH and UseParallelGC enabled by default in the code.

rkennke (approved):
Looks good. (well, except that this size_might_change() is an ugly cludge, but this needs to be addressed elsewhere)

stefank:
Thanks for reviewing!
/integrate

PR #165, opened 2024-04-24
Don't warn when UCOH is on by default

stefank:
When running testing with UCOH enabled by default in the code and non-generational ZGC the JVM prints this warning, which causes multiple failures in the testing. I'd like to propose that we  (at least temporarily) only write this warning if the user explicitly asked for UCOH.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/165/head:pull/165` \
`$ git checkout pull/165`

`$ git checkout pull/165` \
`$ git pull https://git.openjdk.org/lilliput.git pull/165/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 165`

`$ git pr show -t 165`

</details>

<a href="https://git.openjdk.org/lilliput/pull/165.diff">https://git.openjdk.org/lilliput/pull/165.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/165#issuecomment-2074059013)

### DIFF
--- a/src/hotspot/share/runtime/arguments.cpp
+++ b/src/hotspot/share/runtime/arguments.cpp
@@ -2952,12 +2952,6 @@ jint Arguments::finalize_vm_init_args(bool patch_mod_javabase) {
 #endif // CAN_SHOW_REGISTERS_ON_ASSERT
 
 #ifdef _LP64
-  if (UseCompactObjectHeaders && UseZGC && !ZGenerational) {
-    if (FLAG_IS_CMDLINE(UseCompactObjectHeaders)) {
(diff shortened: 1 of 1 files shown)


rkennke (changes requested):
Oh I believe this can be removed altogether. The reason why ZGC had problems was bad interaction/races with displaced headers, but with OMWorld this should no longer be the case.

stefank:
OK. Let me think about that a bit and I'll update the PR later.

rkennke (approved):
Looks good! Thanks!

stefank:
Thanks for reviewing!
/integrate

PR #170, opened 2024-04-24
Fix arrays-equals intrinsic on AArch64

rkennke:
The arrays-equals intrinsic on AArch64 assumes that array elements start at 8-byte-aligned boundary. There are several problems with that:
- I believe it may give wrong results when comparing some junk after the end of the array.
- We may crash when loading beyond the heap boundary.

The proposed fix is to start the comparison at the array-length field. When the array base is unaligned (that is really 4-byte-aligned), then the array-length is at 8-byte-aligned location. And since we want to compare the lengths anyway, we can just as well use word-sized loads to compare the length and first elements in a single step, and elide the separate cmp+branch for the length.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/170/head:pull/170` \
`$ git checkout pull/170`

`$ git checkout pull/170` \
`$ git pull https://git.openjdk.org/lilliput.git pull/170/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 170`

`$ git pr show -t 170`

</details>

<a href="https://git.openjdk.org/lilliput/pull/170.diff">https://git.openjdk.org/lilliput/pull/170.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/170#issuecomment-2075116835)

### DIFF
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -5242,6 +5242,14 @@ address MacroAssembler::arrays_equals(Register a1, Register a2, Register tmp3,
   int length_offset = arrayOopDesc::length_offset_in_bytes();
   int base_offset
     = arrayOopDesc::base_offset_in_bytes(elem_size == 2 ? T_CHAR : T_BYTE);
+  // When the base is not aligned to 8 bytes, then we let
+  // the compare loop include the array length, and skip
+  // the explicit comparison of length.
+  bool is_8aligned = is_aligned(base_offset, BytesPerWord);
+  assert(is_aligned(base_offset, BytesPerWord) || is_aligned(length_offset, BytesPerWord),
+         "base_offset or length_offset must be 8-byte aligned");
+  int start_offset = is_8aligned ? base_offset : length_offset;
+  int extra_length = is_8aligned ? 0 : BytesPerInt / elem_size;
   int stubBytesThreshold = 3 * 64 + (UseSIMDForArrayEquals ? 0 : 16);
 
   assert(elem_size == 1 || elem_size == 2, "must be char or byte");
@@ -5276,10 +5284,15 @@ address MacroAssembler::arrays_equals(Register a1, Register a2, Register tmp3,
     bind(A_IS_NOT_NULL);
     ldrw(cnt1, Address(a1, length_offset));
     ldrw(cnt2, Address(a2, length_offset));
+    if (extra_length != 0) {
+      // Increase loop counter by size of length field.
+      addw(cnt1, cnt1, extra_length);
+      addw(cnt2, cnt2, extra_length);
+    }
     eorw(tmp5, cnt1, cnt2);
     cbnzw(tmp5, DONE);
-    lea(a1, Address(a1, base_offset));
-    lea(a2, Address(a2, base_offset));
+    lea(a1, Address(a1, start_offset));
+    lea(a2, Address(a2, start_offset));
     // Check for short strings, i.e. smaller than wordSize.
     subs(cnt1, cnt1, elem_per_word);
     br(Assembler::LT, SHORT);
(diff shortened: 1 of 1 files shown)


src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -5338,22 +5338,41 @@ address MacroAssembler::arrays_equals(Register a1, Register a2, Register tmp3,
...
     // on most CPUs a2 is still "locked"(surprisingly) in ldrw and it's
     // faster to perform another branch before comparing a1 and a2
     cmp(cnt1, (u1)elem_per_word);
     br(LE, SHORT); // short or same
-    ldr(tmp3, Address(pre(a1, base_offset)));
+    if (is_8aligned) {
+      ldr(tmp3, Address(pre(a1, base_offset)));
+    } else {
+      ldr(tmp3, Address(pre(a1, length_offset)));
+    }
```
xmas92, line 5363: Something like this could avoid the ifs. Also the `const int start_offset = is_8aligned ? base_offset : length_offset;` could be pushed up and shared for the simple implementation, which probably should be fixed as well.
```suggestion
    const int start_offset = is_8aligned ? base_offset : length_offset;
    ldr(tmp3, Address(pre(a1, start_offset)));
```

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -5408,12 +5431,22 @@ address MacroAssembler::arrays_equals(Register a1, Register a2, Register tmp3,
     mov(result, a2);
     b(DONE);
     bind(SHORT);
-    cmp(cnt2, cnt1);
-    br(NE, DONE);
+    if (is_8aligned) {
+      cmp(cnt2, cnt1);
+      br(NE, DONE);
+    }
     cbz(cnt1, SAME);
```
xmas92, line 5438: Was this an optimization? The `cbz` would never be taken if `!is_8aligned `  as it implies `cnt1 > 0`

xmas92 (commented):
This seems resonable. 

But some initial comments.

It seems like the `UseSimpleArrayEquals` above suffers from the unaligned access issues, it seems to be of by default except on Cortex A73. However it should probably be handled as well.

This was also broken for large arrays which assumes that the pointers are 8 byte aligned. The whole over-aligning in there to 16 bytes is just wrong when these are not aligned.
This seems to be the only use of the large array stub, but it seems like all of these have to be evaluated for correctness, have comments and requirements updated. E.g.
https://github.com/openjdk/lilliput/blob/80abb98d830fa26c8381b09c71857f0276c031b9/src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp#L5253-L5254

theRealAph:
> The arrays-equals intrinsic on AArch64 assumes that array elements start at 8-byte-aligned boundary. There are several problems with that:
> 
>     * Doing unaligned loads is likely rather slow.

I doubt that. The design goal for AArch64 was that unaligned loads that don't cross cache-line boundaries shouldn't be penalized.

>     * I believe it may give wrong results when comparing some junk after the end of the array.
> 
>     * We may crash when loading beyond the heap boundary.

The resulting code is really pretty ugly and hard to follow. We should be able to use an array comparison that doesn't have all of those `if (is_8aligned)`, surely? If we need to align, then we just check the first sub-words, then compare as many full-size words as we have, then check any remaining sub-words.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
rkennke: Yes. The idea is that if the main loop also compares the lenghts, then we don't need to explicitely compare them before the main loop and save a bunch of instructions. I removed the special-handling there for now, for simplicity.

xmas92 (approved):
I think this is fine for getting this fixed in Lilliput. T

PR #172, opened 2024-04-27
Prepare for smaller-than-22-bit class pointers

tstuefe:
This PR prepares using arbitrary klass pointer sizes (e.g. 16). It cleans up a few places and corrects comments.

The changes in detail:

- exposes a new function `CompressedKlassPointers::max_encoding_range_size()` that returns the maximum possible size of the encoding range given the current nKlass geometry (e.g. 16 bit klass pointers with a max. shift of 10 bits can encode 64MB of class space). 

- In Metaspace::ergo_initialize(), where we ergo-adjust the CompressedClassSpaceSize, the maximum possible encoding range size flows into this adjustment now. We also print clearer warnings in case the user specifies CCS size explicitly, and we override that decision.

- removed any hard-wiredness of a "max class space size/encoding range size of 4GB" since with smaller geometries that does not hold true anymore. Instead, we now use `CompressedKlassPointers::max_encoding_range_size()`.

- made the requirements on klass_alignment_in_bytes clearer when setting up class space

- removed remnant code (TinyClassPointerShift) left over from development

Note: One still unsolved problem—unsolved in Lilliput as well as upstream—is to correctly limit the compressed class space size in the presence of CDS. Upstream "solves" this by capping CCS size at 3GB, which leaves 1GB for CDS archives. There is no solution if CDS would ever exceed this limit,  and it's a waste of space for CCS. 

In Lilliput, if we limit the class pointer size such that we drastically reduce the klass encoding range size, we need to be better at splitting that klass encoding range between CDS and class space. For example, we could map CDS and then use the remaining space completely for class space. But that would require more serious reshuffling for initialization code, and CDS setup is horrendously complex.

For this patch, if one wants to reduce class pointer size, one may have to disable CDS to run.

Tested: Mac m1, fastdebug, with 32, 22 and 16 bit class pointers. GHAs in process.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/172/head:pull/172` \
`$ git checkout pull/172`

`$ git checkout pull/172` \
`$ git pull https://git.openjdk.org/lilliput.git pull/172/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 172`

`$ git pr show -t 172`

</details>

<a href="https://git.openjdk.org/lilliput/pull/172.diff">https://git.openjdk.org/lilliput/pull/172.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/172#issuecomment-2080405199)

### DIFF
--- a/src/hotspot/share/memory/metaspace.cpp
+++ b/src/hotspot/share/memory/metaspace.cpp
@@ -653,6 +653,8 @@ void Metaspace::ergo_initialize() {
   if (UseCompressedClassPointers) {
     // Adjust size of the compressed class space.
 
+    const size_t res_align = reserve_alignment();
+
     // Let CCS size not be larger than 80% of MaxMetaspaceSize. Note that is
     // grossly over-dimensioned for most usage scenarios; typical ratio of
     // class space : non class space usage is about 1:6. With many small classes,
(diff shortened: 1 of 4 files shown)


tstuefe:
All test errors stem from the fact that we call the JVM with a very small MaxMetaspaceSize, and the CCS size - which is set to 80% of MaxMetaspaceSize - then is down-aligned by root chunk size (16MB). That causes CCS size to be 0 now.

Before, we were aligning CCS size by root chunk size. With Lilliput, that is a bug, since we may have a smaller-than-4GB encoding range, which in turn means that CCS sticks its feet out of the encoding range cover. Upstream this bug is only theoretical since CCS is limited to 3GB and encoding range is 4GB.

tstuefe:
Okay, I now just require the max encoding range to be at least the size of a root chunk (16 MB) and otherwise revert the behavior to before, meaning we round up CCS size to root chunk size. That means that we cannot have < 14 bit  nKlass pointers for now with a 10 bit shift. I think we can live with that.

tstuefe:
Mac x64 errors unrelated.

tstuefe:
keep open bot

rkennke:
I used this change to implement 19-bit-wide class-pointers in my 4-byte-header prototype. Those are the changes on top of this PR that were needed to make that happen:
https://github.com/rkennke/lilliput/commit/2f2ffeadfb566b7ab0eea2aa065140011214b90d

rkennke (approved):
The changes look good to me.

tstuefe:
Thanks @rkennke !

/integrate

PR #174, opened 2024-05-23
OMWorld: reenable all platforms

xmas92:
This reenables all platforms to use OMWorld, and by extension UseCompactObjectHeaders. 

This change simply calls the runtime if a lock is inflated, until port support for OMWorld cache lookup is added.

ARM (32-bit) required no changes as it already always called the runtime when a monitor is inflated.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/174/head:pull/174` \
`$ git checkout pull/174`

`$ git checkout pull/174` \
`$ git pull https://git.openjdk.org/lilliput.git pull/174/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 174`

`$ git pr show -t 174`

</details>

<a href="https://git.openjdk.org/lilliput/pull/174.diff">https://git.openjdk.org/lilliput/pull/174.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/174#issuecomment-2126333174)

### DIFF
--- a/src/hotspot/share/runtime/lightweightSynchronizer.cpp
+++ b/src/hotspot/share/runtime/lightweightSynchronizer.cpp
@@ -1042,6 +1042,16 @@ bool LightweightSynchronizer::quick_enter(oop obj, JavaThread* current, BasicLoc
 
   const markWord mark = obj->mark();
 
+  if (mark.is_unlocked()) {
+    markWord locked_mark = mark.set_fast_locked();
+    if (obj->cas_set_mark(locked_mark, mark) == mark) {
+      // Successfully fast-locked, push object to lock-stack and return.
+      lock_stack.push(obj);
+      current->inc_held_monitor_count();
+      return true;
+    }
+  }
+
   if (mark.has_monitor()) {
     ObjectMonitor* const monitor = current->om_get_from_monitor_cache(obj);
 
--- a/src/hotspot/cpu/ppc/macroAssembler_ppc.cpp
+++ b/src/hotspot/cpu/ppc/macroAssembler_ppc.cpp
@@ -2515,32 +2515,11 @@ void MacroAssembler::compiler_fast_lock_lightweight_object(ConditionRegister fla
   { // Handle inflated monitor.
     bind(inflated);
 
-    // mark contains the tagged ObjectMonitor*.
-    const Register tagged_monitor = mark;
-    const uintptr_t monitor_tag = markWord::monitor_value;
-    const Register owner_addr = tmp2;
+    // OMCache lookup not supported yet. Take slowpath.
 
-    // Compute owner address.
-    addi(owner_addr, tagged_monitor, in_bytes(ObjectMonitor::owner_offset()) - monitor_tag);
-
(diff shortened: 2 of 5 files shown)


coleenp (changes requested):
This is sort of awkward because right now, Lightweight locking is default, so going forward, if we put the OMWorld table on a switch, we'll have to revert this code.  Would it be possible for this Lilliput patch to conditionalize the code on the UseCompactObjectHeaders instead?

xmas92:
> This is sort of awkward because right now, Lightweight locking is default, so going forward, if we put the OMWorld table on a switch, we'll have to revert this code. Would it be possible for this Lilliput patch to conditionalize the code on the UseCompactObjectHeaders instead?

A UseCompactObjectHeaders condition will not work as you can still run LM_LIGHTWEIGHT with -UseCompactObjectHeaders. And that would crash. 

I was thinking this could be done when introducing the OMWorld flag. (And simply reintroduce the code this removes).

The alternative is that I wait with this patch until you have introduced the UseOMWorld flag. And I can then update this to instead of removing code condition the runtime call on the OMWorld flag.

src/hotspot/share/runtime/lightweightSynchronizer.cpp
```diff
@@ -1042,6 +1042,16 @@ bool LightweightSynchronizer::quick_enter(oop obj, JavaThread* current, BasicLoc
 
   const markWord mark = obj->mark();
 
+  if (mark.is_unlocked()) {
```
rkennke, line 1045: Was that not needed before? Or why is this added now?

rkennke (commented):
Hi Axel, this mostly looks good, I have just one question.

src/hotspot/share/runtime/lightweightSynchronizer.cpp
coleenp: Also wondering this.  We come here if the object is locked via slow path, but can it become unlocked only for the platforms that come don't check the ObjectMonitor and go slow path?

coleenp:
The changes for this PR are also in this PR:  https://github.com/openjdk/lilliput/pull/181/   @rkennke  can you have a look at this? Thanks.

src/hotspot/share/runtime/lightweightSynchronizer.cpp
xmas92: While it is possible that someone has unlocked/deflated by the time we get here. This was really thought as a bandaid for `x86_32`.  Which now skips fast locking due to lack of a thread register in some paths. 

I think the whole `quick_enter` should be reevaluated in case when we get here from C2. @coleenp has been running some experiments where it is removed completely.

I can see a solution where C1 and C2 have two different entry points, where C1 does the `quick_enter` (and only fast_locking on `x86_32`) while the C2 entry point skips it. As it only gets here in a true slow path (except for `arm32`)

xmas92:
Closing this as it was merged as a part of #181

PR #175, opened 2024-05-23
OMWorld: Decouple deflation and table sizing

xmas92:
The change reverts all changes to deflation and moves the resizing of the OMWorld ConcurrentHashTable to the service thread. Using a similar logic to how we resize the Symbol- and StringTables. 

The option to shrink the table is taken out and can be reintroduced at a later date as an enhancement. To do it correctly the interactions with deflation needs to be figured out.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/175/head:pull/175` \
`$ git checkout pull/175`

`$ git checkout pull/175` \
`$ git pull https://git.openjdk.org/lilliput.git pull/175/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 175`

`$ git pr show -t 175`

</details>

<a href="https://git.openjdk.org/lilliput/pull/175.diff">https://git.openjdk.org/lilliput/pull/175.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/175#issuecomment-2126341188)


src/hotspot/share/runtime/serviceThread.cpp
```diff
@@ -196,6 +199,10 @@ void ServiceThread::service_thread_entry(JavaThread* jt, TRAPS) {
     if (jvmti_tagmap_work) {
       JvmtiTagMap::flush_all_object_free_events();
     }
+
+    if (omworldtable_work) {
+      LightweightSynchronizer::resize_table(jt);
+    }
```
coleenp, line 205: The monitor deflation thread seems like it would be a better place to do the OMWorld table resizing, because it also does cleaning that might be a response to monitor deflation.  The ServiceThread does a lot, and the OMWorld table might notice the stall.  The other thread is also a service thread in a way, so I think this could be moved there instead.

coleenp: And when we do add shrinking, there'll be interactions with things that the monitor deflation thread is doing.

coleenp (commented):
This does simplify the table to act like the symbol and string tables, until we find we need to do something more complicated.

coleenp (approved):
The interaction with the OM deflation thread isn't straightforward so it seems better to have this in the ServiceThread for now.  Thanks for explaining this @xmas92

PR #177, opened 2024-05-23
OMWorld: Spin Changes

xmas92:
The fast lock spinning uses `sched_yield` which tends to be discouraged for spin locking code. Instead only use `SpinPause` with exponential backoff. Where after each failed CAS wait for exponentially more time until trying again in an attempt to reduce cache contention. 

This change also makes the spinning aware of safepoints, and tries to fast track the execution to next poll, which is either when successfully locked (VM backedge transition) or when going into blocked to enter the ObjectMonitor.

Have not removed `OMSpins` yet, as the exact value is not determined yet. It may have to be platform specific as `SpinPause` have different characteristics on different hardware. OMSpins is the number of fast lock, with each attempt spinning for twice as much as the last, so the total number of spins are on the order of O(2^OMSpins). It will probably land somewhere on the range of 7-14 (128 -16384 spins)

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/177/head:pull/177` \
`$ git checkout pull/177`

`$ git checkout pull/177` \
`$ git pull https://git.openjdk.org/lilliput.git pull/177/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 177`

`$ git pr show -t 177`

</details>

<a href="https://git.openjdk.org/lilliput/pull/177.diff">https://git.openjdk.org/lilliput/pull/177.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/177#issuecomment-2126369051)

### DIFF
--- a/src/hotspot/share/runtime/lightweightSynchronizer.cpp
+++ b/src/hotspot/share/runtime/lightweightSynchronizer.cpp
@@ -570,10 +570,10 @@ bool LightweightSynchronizer::fast_lock_spin_enter(oop obj, JavaThread* current,
   LockStack& lock_stack = current->lock_stack();
 
   markWord mark = obj->mark();
-  const bool try_spin = observed_deflation || !mark.has_monitor();
+  const auto try_spin = [&]() { return observed_deflation || !mark.has_monitor(); };
   // Always attempt to lock once even when safepoint synchronizing.
   bool should_process = false;
-  for (int i = 0; try_spin && !should_process && i < log_spin_limit; i++) {
+  for (int i = 0; try_spin() && !should_process && i < log_spin_limit; i++) {
     // Spin with exponential backoff.
     const int total_spin_count = 1 << i;
     const int inner_spin_count = MIN2(1 << log_min_safepoint_check_interval, total_spin_count);
--- a/src/hotspot/share/runtime/lightweightSynchronizer.cpp
+++ b/src/hotspot/share/runtime/lightweightSynchronizer.cpp
@@ -634,7 +634,7 @@ void LightweightSynchronizer::enter(Handle obj, BasicLock* lock, JavaThread* cur
   }
 
   // Will spin with exponential backoff with an accumulative O(2^spin_limit) spins.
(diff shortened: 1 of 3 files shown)


src/hotspot/share/runtime/lightweightSynchronizer.cpp
```diff
@@ -628,16 +633,33 @@ void LightweightSynchronizer::enter(Handle obj, BasicLock* lock, JavaThread* cur
...
+        if (should_process) {
+          // Stop spinning for safepoint.
+          break;
+        }
+        for (int inner = 1; inner < inner_spin_count; inner++) {
+          SpinPause();
+        }
+      }
+
+      mark = obj()->mark();
```
coleenp, line 662: Is this the spinning to delay creating an ObjectMonitor for this lock as long as possible?  Or if deflation is observed?  This seems like it'd slow down the contended case, like xalan where we want a lot of the threads to quickly park.  Or is this spinning because deflation is observed (like the comment says above)?

Can you make 647-660 a function above this so it doesn't distract from the logic of the rest of enter so much?

xmas92: > Is this the spinning to delay creating an ObjectMonitor for this lock as long as possible? Or if deflation is observed? This seems like it'd slow down the contended case, like xalan where we want a lot of the threads to quickly park. Or is this spinning because deflation is observed (like the comment says above)?

There are two spinning loops here. The outer which uses the SpinYield which is only to do with deflation. And the inner when the object is fast_lock. But the first_time variable has a bad name. And with the new exponential backoff changes the spinning should be interrupted if inflation is observed.


> Can you make 647-660 a function above this so it doesn't distract from the logic of the rest of enter so much?

Good suggestion. Becomes much cleared. 

Extracted the spinning logic, rename the variables and cleaned up the loop condition.

coleenp (approved):
I like the comments.  Thank you!

PR #178, opened 2024-05-23
OMWorld: Cleanups

xmas92:
This contains a handful of miscellaneous cleanups. Removing dead code, fixes/cleanups todos, cleanup logging and removing all special handling of `x86_32` not having a thread register / lacking registers.

### DIFF
--- a/src/hotspot/cpu/aarch64/c2_MacroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/c2_MacroAssembler_aarch64.cpp
@@ -236,7 +236,7 @@ void C2_MacroAssembler::fast_lock_lightweight(Register obj, Register box, Regist
   // Finish fast lock unsuccessfully. MUST branch to with flag == NE
   Label slow_path;
 
-  // Clear box. TODO[OMWorld]: Is this necessary? May also defer this to not write twice.
+  // Clear cache in case fast locking succeeds.
   str(zr, Address(box, BasicLock::object_monitor_cache_offset_in_bytes()));
 
   if (DiagnoseSyncOnValueBasedClasses != 0) {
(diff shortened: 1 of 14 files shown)


src/hotspot/cpu/aarch64/c2_MacroAssembler_aarch64.cpp
```diff
@@ -498,9 +498,8 @@ void C2_MacroAssembler::fast_unlock_lightweight(Register obj, Register box, Regi
 
       if (OMCacheHitRate) increment(Address(rthread, JavaThread::unlock_lookup_offset()));
       ldr(t1_monitor, Address(box, BasicLock::object_monitor_cache_offset_in_bytes()));
-      // TODO: Cleanup these constants (with an enum and asserts)
-      cmp(t1_monitor, (uint8_t)2);
-      // Non symmetrical, take slow path monitor == 0 or 1, 0 and 1 < 2, both LS and NE
+      // null check with Flags == NE, no valid pointer below alignof(ObjectMonitor*)
+      cmp(t1_monitor, checked_cast<uint8_t>(alignof(ObjectMonitor*)));
```
coleenp, line 502: I don't know what this means.

coleenp (approved):
These cleanups look good.

src/hotspot/cpu/aarch64/c2_MacroAssembler_aarch64.cpp
xmas92: Maybe I should say `valid ObjectMonitor* check` instead of `null check`. 

The reason that the check used to be against `2` is because some iterations of the code have signalled information in the lowest bit. 

Currently it could just be a check agains `1` which would check for `nullptr` while also setting Flags to NE. 

The check is simply that based on the fact that no valid ObjectMonitor* except nullptr can exist below `alignof(ObjectMonitor*)`.

PR #181, opened 2024-05-31
Lilliput om world

coleenp:
Added a diagnostic option UseObjectMonitorTable to maintain performance of LM_LIGHTWEIGHT locking which is now defaulted to on in mainline.

Incorporated Axel's patch to support other platforms with UseObjectMonitorTable (branch to slow path).

Tested tier 1-4 on x86.  Fails both versions of this test, but not locally:  runtime/cds/TestDefaultArchiveLoading.java.

tier1 aarch64 in progress.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/181/head:pull/181` \
`$ git checkout pull/181`

`$ git checkout pull/181` \
`$ git pull https://git.openjdk.org/lilliput.git pull/181/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 181`

`$ git pr show -t 181`

</details>

<a href="https://git.openjdk.org/lilliput/pull/181.diff">https://git.openjdk.org/lilliput/pull/181.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/181#issuecomment-2148528996)


src/hotspot/share/runtime/synchronizer.cpp
```diff
@@ -415,14 +410,34 @@ bool ObjectSynchronizer::quick_enter(oop obj, JavaThread* current,
...
+    LockStack& lock_stack = current->lock_stack();
+    if (lock_stack.is_full()) {
+      // Always go into runtime if the lock stack is full.
+      return false;
+    }
+
+    if (lock_stack.try_recursive_enter(obj)) {
+      // Recursive lock successful.
+      current->inc_held_monitor_count();
+      return true;
```
xmas92, line 427: This needs a CacheSetter

```suggestion
      CacheSetter cache_setter(current, lock);
  
      // Recursive lock successful.
      current->inc_held_monitor_count();
      return true;
```

xmas92 (commented):
I like this change and the restructuring. 👍

src/hotspot/share/runtime/synchronizer.cpp
coleenp: Ok, yes, I missed that the fast lock stack case set the monitor.  I was trying to keep the CacheSetter contained in lightweightSynchronizer.cpp.  I'll do some more refactoring.  Thanks for pointing this out.

coleenp: Oh now I see.  If we do a recursive lock, we need to clear_object_monitor_cache for some reason I haven't figured out yet.

xmas92: Yeah I am not 100% happy with the CacheSetter, it at least needs better documentation. The issue is an uninitialised value will (more than likely) look like a ObjectMonitor* to the cache lookup code. So it will break if this thread observers that it is inflate (by for example calling wait) then it will crash when exiting.

It aims to uphold the simple invariant that after ObjectSynchronize::enter the BasicLock contains either the ObjectMonitor locked on or nullptr.

xmas92:
I did a more thorough review after the rebase. There are a couple of things I found. Because some changes are in other files I created a review patch below. I opened a pr to your fork if you want to merge it in and edit it. https://github.com/coleenp/lilliput/pull/1

xmas92 (approved):
lgtm.

coleenp:
Thank you Axel.
/integrate

PR #183, opened 2024-06-20
OMWorld: Removal of flags, general cleanups

xmas92:
Cleanups in preparation for opening up a PR to mainline for [JDK-8315884](https://bugs.openjdk.org/browse/JDK-8315884).

@coleenp I incorporated #182 into this.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/183/head:pull/183` \
`$ git checkout pull/183`

`$ git checkout pull/183` \
`$ git pull https://git.openjdk.org/lilliput.git pull/183/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 183`

`$ git pr show -t 183`

</details>

<a href="https://git.openjdk.org/lilliput/pull/183.diff">https://git.openjdk.org/lilliput/pull/183.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/183#issuecomment-2180503447)

### DIFF
--- a/src/hotspot/cpu/x86/sharedRuntime_x86.cpp
+++ b/src/hotspot/cpu/x86/sharedRuntime_x86.cpp
@@ -62,9 +62,11 @@ void SharedRuntime::inline_check_hashcode_from_object_header(MacroAssembler* mas
 
 
   if (LockingMode == LM_LIGHTWEIGHT) {
-    // check if monitor
-    __ testptr(result, markWord::monitor_value);
-    __ jcc(Assembler::notZero, slowCase);
+    if (!UseObjectMonitorTable) {
+      // check if monitor
+      __ testptr(result, markWord::monitor_value);
+      __ jcc(Assembler::notZero, slowCase);
+    }
   } else {
     // check if locked
     __ testptr(result, markWord::unlocked_value);
--- a/src/hotspot/share/runtime/globals.hpp
+++ b/src/hotspot/share/runtime/globals.hpp
@@ -1977,22 +1977,12 @@ const int ObjectAlignmentInBytes = 8;
           "With Lightweight Locking mode, use a table to record inflated "  \
           "monitors rather than the first word of the object.")             \
                                                                             \
-  product(bool, OMC2UnrollCacheLookupLoopTail, true, "")                    \
-                                                                            \
-  product(int, OMC2UnrollCacheEntries, 0, "")                               \
-          range(0, OMCache::CAPACITY)                                       \
-                                                                            \
-  product(int, OMCacheSize, 8, "")                                          \
-          range(0, OMCache::CAPACITY)                                       \
-                                                                            \
-  product(int, OMSpins, 13,                                                 \
+  product(int, LightweightFastLockingSpins, 13, DIAGNOSTIC,                 \
           "Specifies the number of time lightweight fast locking will "     \
           "attempt to CAS the markWord before inflating. Between each "     \
           "CAS it will spin for exponentially more time, resulting in "     \
-          "a total number of spins on the order of O(2^OMSpins)")           \
+          "a total number of spins on the order of O(2^value)")             \
           range(1, 30)                                                      \
-                                                                            \
-  product(bool, OMCacheHitRate, false, "")                                  \
                                                                             \
   product(uint, TrimNativeHeapInterval, 0,                                  \
           "Interval, in ms, at which the JVM will trim the native heap if " \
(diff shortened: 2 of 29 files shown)


xmas92:
I will do a rebase on JDK-24 next week after getting these in. 

Already did one earlier this week. But I felt like it was better to get this OMWorld cleanups in. (I believe @coleenp is about done with the initial pre-review of it.) But one not on the next rebase, as parallel full GC has been changed to work more like G1 it will also require adaption with to use the Alt Full GC forwarding. E.g. from my earlier rebase ae4715b7d7ce6297d939b4eba5525581765a7b95

src/hotspot/share/utilities/vmError.cpp
```diff
@@ -590,9 +590,6 @@ static void report_vm_version(outputStream* st, char* buf, int buflen) {
 #endif
                  UseCompressedOops ? ", compressed oops" : "",
                  UseCompressedClassPointers ? ", compressed class ptrs" : "",
-                 LockingMode == LM_MONITOR ? ", lm_monitors" :
-                 LockingMode == LM_LEGACY ? ", lm_legacy" :
-                 LockingMode == LM_LIGHTWEIGHT ? ", lm_lightweight" : "",
```
coleenp, line 595: Without this, I guess the only way to know which locking mode is on is via. the command line in the hs_err file.  It didn't really belong here, but maybe we want to find somewhere else for it.  Not with this change.  It's good to clean this up for this change.

src/hotspot/share/runtime/objectMonitor.cpp
```diff
@@ -296,10 +296,10 @@ void ObjectMonitor::ClearSuccOnSuspend::operator()(JavaThread* current) {
   }
 }
 
-#define assert_mark_word_concistency()                                                 \
+#define assert_mark_word_concistency()                                         \
```
coleenp, line 299: While you're here. The spelling is consistency.

src/hotspot/share/runtime/globals.hpp
```diff
@@ -1977,23 +1977,13 @@ const int ObjectAlignmentInBytes = 8;
...
-  product(bool, OMC2UnrollCacheLookupLoopTail, true, "")                    \
-                                                                            \
-  product(int, OMC2UnrollCacheEntries, 0, "")                               \
-          range(0, OMCache::CAPACITY)                                       \
-                                                                            \
-  product(int, OMCacheSize, 8, "")                                          \
-          range(0, OMCache::CAPACITY)                                       \
-                                                                            \
-  product(int, OMSpins, 13,                                                 \
+  product(int, LightweightFastLockingSpins, 13, DIAGNOSTIC,                 \
```
coleenp, line 1980: Good.  I like the new name.

src/hotspot/cpu/x86/sharedRuntime_x86.cpp
```diff
@@ -62,9 +62,11 @@ void SharedRuntime::inline_check_hashcode_from_object_header(MacroAssembler* mas
...
 
   if (LockingMode == LM_LIGHTWEIGHT) {
-    // check if monitor
-    __ testptr(result, markWord::monitor_value);
-    __ jcc(Assembler::notZero, slowCase);
+    if (!UseObjectMonitorTable) {
+      // check if monitor
+      __ testptr(result, markWord::monitor_value);
+      __ jcc(Assembler::notZero, slowCase);
+    }
```
coleenp, line 69: Did I miss this one?

coleenp (approved):
The cleanups look really good.  Can you fix the spelling of consistency?

src/hotspot/share/utilities/vmError.cpp
xmas92: We print flags at the end which are non-default. But we do not know if someone changed them in a build. 
Should probably fix this somehow. 

Also printing all flag values, not just non-default ones would help.

src/hotspot/cpu/x86/sharedRuntime_x86.cpp
xmas92: We did the C2 one. This is the native wrapper one. We both overlooked it.

coleenp: I see.  We missed an optimization here.

coleenp (approved):

coleenp (approved):

xmas92:
Thanks for the review.
/integrate

PR #187, opened 2024-07-01
Add some inlining and avoid CAS for quick_enter for lightweight locking.

coleenp:
This is sort of a simple change to inline the code that decides to go to LightweightSynchronizer in case the call has some negative effects for performance sensitive code.  Avoiding the CAS in try_enter was the most helpful.  With perf, I found that the CAS had a longer stall than with the code without the OM world refactoring.  The OM world table is off for this comparison.
Tested with tier1-4, including some local changes to run tier1 tests with -XX:+UseObjectMonitorTable.

---------
### Progress

### Reviewers
 * [Axel Boldt-Christmas](https://openjdk.org/census#aboldtch) (@xmas92 - Committer) ⚠️ Review applies to [fb00e8dd](https://git.openjdk.org/lilliput/pull/187/files/fb00e8ddfe223082abb1e260aff3cbbb4de9ffd0)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/187/head:pull/187` \
`$ git checkout pull/187`

`$ git checkout pull/187` \
`$ git pull https://git.openjdk.org/lilliput.git pull/187/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 187`

`$ git pr show -t 187`

</details>

<a href="https://git.openjdk.org/lilliput/pull/187.diff">https://git.openjdk.org/lilliput/pull/187.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/187#issuecomment-2200957721)

### DIFF
--- a/src/hotspot/share/runtime/lightweightSynchronizer.cpp
+++ b/src/hotspot/share/runtime/lightweightSynchronizer.cpp
@@ -575,12 +575,31 @@ public:
   }
 };
 
-bool LightweightSynchronizer::fast_lock_spin_enter(oop obj, JavaThread* current, bool observed_deflation) {
-  // Will spin with exponential backoff with an accumulative O(2^spin_limit) spins.
-  const int log_spin_limit = os::is_MP() || !UseObjectMonitorTable ? LightweightFastLockingSpins : 1;
-  const int log_min_safepoint_check_interval = 10;
+inline bool LightweightSynchronizer::check_unlocked(oop obj, LockStack& lock_stack, JavaThread* current) {
+  markWord mark = obj->mark();
+  while (mark.is_unlocked()) {
+    ensure_lock_stack_space(current);
+    assert(!lock_stack.is_full(), "must have made room on the lock stack");
+    assert(!lock_stack.contains(obj), "thread must not already hold the lock");
+    // Try to swing into 'fast-locked' state.
+    markWord locked_mark = mark.set_fast_locked();
+    markWord old_mark = mark;
+    mark = obj->cas_set_mark(locked_mark, old_mark);
+    if (old_mark == mark) {
+      // Successfully fast-locked, push object to lock-stack and return.
+      lock_stack.push(obj);
+      return true;
+    }
+  }
+  return false;
+}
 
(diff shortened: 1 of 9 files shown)


src/hotspot/share/runtime/objectMonitor.cpp
```diff
@@ -382,22 +382,38 @@ bool ObjectMonitor::enter_for(JavaThread* locking_thread) {
 }
 
 bool ObjectMonitor::try_enter(JavaThread* current) {
-  void* cur = try_set_owner_from(nullptr, current);
-  if (cur == nullptr) {
-    assert(_recursions == 0, "invariant");
-    return true;
-  }
+  if (LockingMode == LM_LIGHTWEIGHT) {
```
xmas92, line 385: Could probably just do this for all LockingModes.

xmas92 (approved):

src/hotspot/share/runtime/lightweightSynchronizer.cpp
```diff
@@ -575,13 +575,32 @@ class LightweightSynchronizer::VerifyThreadState {
   }
 };
 
-bool LightweightSynchronizer::fast_lock_spin_enter(oop obj, JavaThread* current, bool observed_deflation) {
+inline bool LightweightSynchronizer::check_unlocked(oop obj, LockStack& lock_stack, JavaThread* current) {
```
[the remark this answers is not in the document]
coleenp: I like fast_lock_try_enter - that looks good in the code and describes the situation.

src/hotspot/share/runtime/objectMonitor.cpp
coleenp: Yes, you convinced me that reading the owner for legacy mode doesn't need to be done as the return from the CAS because it can change anyway.  If the owner is the stack lock, which the code is testing for, it will never change here so it's ok to read it again.

coleenp (commented):
Thank you for your comments, Axel.

coleenp:
Thanks Axel for the review.
/integrate

PR #193, opened 2025-01-21
8347710: [Lilliput] Implement 4 byte headers

rkennke:
Packs/reduces header bits to just 4 bytes.

This reduces the number of Klass* bits to 19 bits, which allows for ~500,000 classes.

Testing:

- tier1 (-UCOH)
- tier2 (-UCOH)
- tier1 (+UCOH)
- tier2 (+UCOH)

---------
### Progress

### Issue
 * [JDK-8347710](https://bugs.openjdk.org/browse/JDK-8347710): [Lilliput] Implement 4 byte headers (**Enhancement** - P4)

### Reviewers
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - **Reviewer**) ⚠️ Review applies to [a1c5b29f](https://git.openjdk.org/lilliput/pull/193/files/a1c5b29f2c21a7315a3d8699a00365eb118a3b51)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/193/head:pull/193` \
`$ git checkout pull/193`

`$ git checkout pull/193` \
`$ git pull https://git.openjdk.org/lilliput.git pull/193/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 193`

`$ git pr show -t 193`

</details>

<a href="https://git.openjdk.org/lilliput/pull/193.diff">https://git.openjdk.org/lilliput/pull/193.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/193#issuecomment-2604708452)
</details>


tstuefe (approved):

src/hotspot/cpu/aarch64/aarch64.ad
```diff
@@ -6708,21 +6711,22 @@ instruct loadNKlass(iRegNNoSp dst, memory4 mem)
   ins_pipe(iload_reg_mem);
 %}
 
-instruct loadNKlassCompactHeaders(iRegNNoSp dst, memory4 mem)
+instruct loadNKlassCompactHeaders(iRegNNoSp dst, memory_noindex mem)
 %{
   match(Set dst (LoadNKlass mem));
   predicate(!needs_acquiring_load(n) && UseCompactObjectHeaders);
```
theRealAph, line 6717: Is there ever a time when we have an acquiring load on a narrow klass pointer?
Also, do we ever generate code which loads a narrow klass pointer from something other than an Object? Maybe there's a use somewhere of `LoadNKlass` from somewhere other than `obj-start + klass_offset_in_bytes`?

adinn: > Is there ever a time when we have an acquiring load on a narrow klass pointer?

That's a good question! This predicate was added to handle volatile instance field loads (there is a similar set of cases for volatile stores). I believe I unthinkingly allowed for the possibility of the acquire case for Klass and NKlass loads (likewise stores) when I first added the rules that used the predicate (I think I just blindly changed all the rules that did a load/store). At that stage the test was done by traversing the graph but even then it was probably always going to fail for a Klass load/store.

In the current implementation the predicate is implemented by checking field _mo of the relevant LoadNode which is set at node created does not appear ever to get reset. Looking at the opto sources it appears that LoadKlassNode and LoadNKlassNode only ever seem to be created created with _mo == unordered. So, I think the Klass laod/store rules whcih check for acquire/release semantics can be dropped.

It would be worth seeking confirmation of that from @rwestrel who was the one who changed the implementation so it relied on the `_mo` fields in `Load/StoreNode`.

rkennke: > Is there ever a time when we have an acquiring load on a narrow klass pointer?

I don't think so, but I copied it from the original loadNKlass code.

> Also, do we ever generate code which loads a narrow klass pointer from something other than an Object? Maybe there's a use somewhere of `LoadNKlass` from somewhere other than `obj-start + klass_offset_in_bytes`?

No, I don't think so, I've checked the C2 code for that. Any paths that load from non-objects are using LoadKlass.

That said, I would love to have a cleaner way to load the (narrow) Klass* out of an object. There has been some discussion around that problem earlier: https://bugs.openjdk.org/browse/JDK-8340453

src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
```diff
@@ -710,20 +710,24 @@ int BarrierSetC2::arraycopy_payload_base_offset(bool is_array) {
...
     assert(UseCompressedClassPointers, "");
-    assert(!UseCompactObjectHeaders, "");
     if (is_array) {
       // Exclude length to copy by 8 bytes words.
       base_off += sizeof(int);
     } else {
-      // Include klass to copy by 8 bytes words.
-      base_off = instanceOopDesc::klass_offset_in_bytes();
+      if (UseCompactObjectHeaders) {
+        base_off = 0; /* FIXME */
```
JohnTortugo, line 724: Should this have been fixed before merge?

src/hotspot/share/jvmci/jvmciCompilerToVM.cpp
```diff
@@ -451,7 +451,7 @@ C2V_VMENTRY_NULL(jobject, getResolvedJavaType0, (JNIEnv* env, jobject, jobject b
 
   const char* base_desc = nullptr;
   JVMCIKlassHandle klass(THREAD);
-  if (offset == oopDesc::klass_offset_in_bytes()) {
+  if (offset == 1 /*oopDesc::klass_offset_in_bytes()*/) {
```
JohnTortugo, line 454: @rkennke I'm seeing a crash internally in Graal and I'm suspecting that this line may be the culprit. The error is:

```
Caused by: java.lang.IllegalArgumentException: Unexpected arguments: Ljdk/vm/ci/hotspot/DirectHotSpotObjectConstantImpl; 8 true
	at jdk.internal.vm.ci/jdk.vm.ci.hotspot.CompilerToVM.getResolvedJavaType0(Native Method)
	at jdk.internal.vm.ci/jdk.vm.ci.hotspot.CompilerToVM.getResolvedJavaType(CompilerToVM.java:1020)
...
```
I'm wondering if the commented part of the condition was supposed to be uncommented at some point.

PR #195, opened 2025-03-14
8347711: [Lilliput] Parallel GC support for compact identity hashcode

rkennke:
The Parallel GC does not yet support Lilliput 2 until now. The big problem has been that the Parallel Full GC is too rigid with respect to object sizes, and we could not make it work with compact identity hashcode, which requires that objects can grow during GC.

The PR implements an alternative full GC for Parallel GC, which is more flexible. The algorithm mostly follows G1 and Shenandoah, with the difference that it creates temporary 'regions' (because Parallel GC does not use heap regions), with boundaries such that no object crosses region boundaries, and then after GC fill any gaps at end of regions with dummy objects.

The implementation has a special 'serial' mode, which sets up only 4 regions which exactly match the 4 heap spaces (old, eden, from, to), and performs the forwarding and compaction phases serially to achieve perfect compaction at the expense of performance. (The marking and adjust-refs phases will still be done with parallel workers).

I've run the micro benchmarks for systemgc, there seem to be only minor differences, and looks mostly like a few milliseconds offset in the new implementation:

Baseline Full GC:
```
AllDead.gc                        ss   25   31.120 ±  0.447  ms/op
AllLive.gc                        ss   25   83.655 ±  2.238  ms/op
DifferentObjectSizesArray.gc      ss   25  179.725 ±  1.171  ms/op
DifferentObjectSizesHashMap.gc    ss   25  186.011 ±  1.409  ms/op
DifferentObjectSizesTreeMap.gc    ss   25   65.668 ±  3.333  ms/op
HalfDeadFirstPart.gc              ss   25   64.862 ±  0.696  ms/op
HalfDeadInterleaved.gc            ss   25   67.764 ±  3.139  ms/op
HalfDeadInterleavedChunks.gc      ss   25   59.160 ±  1.667  ms/op
HalfDeadSecondPart.gc             ss   25   66.210 ±  1.167  ms/op
HalfHashedHalfDead.gc             ss   25   69.584 ±  2.276  ms/op
NoObjects.gc                      ss   25   18.462 ±  0.270  ms/op
OneBigObject.gc                   ss   25  587.425 ± 27.493  ms/op
```

New Parallel Full GC:

```
AllDead.gc                        ss   25   39.891 ±  0.461  ms/op
AllLive.gc                        ss   25   87.898 ±  1.940  ms/op
DifferentObjectSizesArray.gc      ss   25  184.109 ±  0.795  ms/op
DifferentObjectSizesHashMap.gc    ss   25  189.620 ±  2.236  ms/op
DifferentObjectSizesTreeMap.gc    ss   25   69.915 ±  3.308  ms/op
HalfDeadFirstPart.gc              ss   25   70.664 ±  0.804  ms/op
HalfDeadInterleaved.gc            ss   25   71.318 ±  1.583  ms/op
HalfDeadInterleavedChunks.gc      ss   25   65.050 ±  1.827  ms/op
HalfDeadSecondPart.gc             ss   25   70.964 ±  0.878  ms/op
HalfHashedHalfDead.gc             ss   25   72.506 ±  1.617  ms/op
NoObjects.gc                      ss   25   23.809 ±  0.494  ms/op
OneBigObject.gc                   ss   25  403.461 ± 27.079  ms/op
```

Testing:
 - [x] tier1 (+UseParallelGC +UCOH)
 - [x] tier2 (+UseParallelGC +UCOH)
 - [x] hotspot_gc (+UCOH)

---------
### Progress

### Issue
 * [JDK-8347711](https://bugs.openjdk.org/browse/JDK-8347711): [Lilliput] Parallel GC support for compact identity hashcode (**Enhancement** - P4)

### Reviewers
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/195/head:pull/195` \
`$ git checkout pull/195`

`$ git checkout pull/195` \
`$ git pull https://git.openjdk.org/lilliput.git pull/195/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 195`

`$ git pr show -t 195`

</details>

<a href="https://git.openjdk.org/lilliput/pull/195.diff">https://git.openjdk.org/lilliput/pull/195.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/195#issuecomment-2725279487)
</details>

### DIFF
--- a/src/hotspot/share/cds/heapShared.cpp
+++ b/src/hotspot/share/cds/heapShared.cpp
@@ -589,7 +589,7 @@ static void copy_java_mirror_hashcode(oop orig_mirror, oop scratch_m) {
         scratch_m->initialize_hash_if_necessary(orig_mirror, orig_klass, mark);
       } else {
         assert(mark.is_hashed_expanded(), "must be hashed & moved");
-        int offset = orig_klass->hash_offset_in_bytes(orig_mirror, mark);
+        int offset = orig_klass->hash_offset_in_bytes(orig_mirror);
         assert(offset >= 8, "hash offset must not be in header");
         scratch_m->int_field_put(offset, (jint) src_hash);
         scratch_m->set_mark(mark);
--- a/src/hotspot/share/cds/heapShared.cpp
+++ b/src/hotspot/share/cds/heapShared.cpp
@@ -589,7 +589,7 @@ static void copy_java_mirror_hashcode(oop orig_mirror, oop scratch_m) {
         scratch_m->initialize_hash_if_necessary(orig_mirror, orig_klass, mark);
       } else {
         assert(mark.is_hashed_expanded(), "must be hashed & moved");
-        int offset = orig_klass->hash_offset_in_bytes(orig_mirror);
+        int offset = orig_klass->hash_offset_in_bytes(orig_mirror, mark);
         assert(offset >= 8, "hash offset must not be in header");
         scratch_m->int_field_put(offset, (jint) src_hash);
         scratch_m->set_mark(mark);
--- a/src/hotspot/share/ci/ciInstanceKlass.hpp
+++ b/src/hotspot/share/ci/ciInstanceKlass.hpp
@@ -297,7 +297,7 @@ public:
   GrowableArray<ciInstanceKlass*>* transitive_interfaces() const;
 
   int hash_offset_in_bytes() const {
-    return get_instanceKlass()->hash_offset_in_bytes(nullptr);
+    return get_instanceKlass()->hash_offset_in_bytes(nullptr, markWord(0));
   }
 
   // Replay support
--- a/src/hotspot/share/ci/ciInstanceKlass.hpp
+++ b/src/hotspot/share/ci/ciInstanceKlass.hpp
@@ -297,7 +297,7 @@ public:
   GrowableArray<ciInstanceKlass*>* transitive_interfaces() const;
 
   int hash_offset_in_bytes() const {
-    return get_instanceKlass()->hash_offset_in_bytes(nullptr, markWord(0));
+    return get_instanceKlass()->hash_offset_in_bytes(nullptr);
   }
 
   // Replay support
--- a/src/hotspot/share/gc/parallel/parMarkBitMap.inline.hpp
+++ b/src/hotspot/share/gc/parallel/parMarkBitMap.inline.hpp
@@ -116,7 +116,7 @@ inline HeapWord* ParMarkBitMap::find_obj_beg(HeapWord* beg, HeapWord* end) const
 inline HeapWord* ParMarkBitMap::find_obj_beg_reverse(HeapWord* beg, HeapWord* end) const {
   const idx_t beg_bit = addr_to_bit(beg);
   const idx_t end_bit = addr_to_bit(end);
-  const idx_t res_bit = _beg_bits.find_last_set_bit_aligned_left(beg_bit, end_bit);
+  const idx_t res_bit = _beg_bits.find_last_set_bit(beg_bit, end_bit);
   return bit_to_addr(res_bit);
 }
 
--- a/src/hotspot/share/gc/parallel/parallelScavengeHeap.cpp
+++ b/src/hotspot/share/gc/parallel/parallelScavengeHeap.cpp
@@ -29,6 +29,7 @@
 #include "gc/parallel/psAdaptiveSizePolicy.hpp"
 #include "gc/parallel/psMemoryPool.hpp"
 #include "gc/parallel/psParallelCompact.inline.hpp"
+#include "gc/parallel/psParallelCompactNew.inline.hpp"
 #include "gc/parallel/psPromotionManager.hpp"
 #include "gc/parallel/psScavenge.hpp"
 #include "gc/parallel/psVMOperations.hpp"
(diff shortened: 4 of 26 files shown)


zhengyu123:
Can you articulate why object cannot grow?

rkennke:
> Can you articulate why object cannot grow?

Yes.
The way PSParallelCompact currently works: it collects liveness information during the marking phase. At the end of it, we know precisely how many words are live in each reagion. In this phase, we also determine where objects cross region boundaries. Then, during the summary phase, it uses this information to determine which source regions get compacted into which target regions. It also determines the addresses at which a region gets split - that is the part below the split-point gets compacted into one region, and the part above the split point gets compacted into another region. Then, during the forwarding phase, all those infos are used to calculate the target addresses for each object in parallel. It is now possible to divide the forwarding work across several GC worker threads because we know exactly where each region gets compacted into.

The problem is that, with compact identity hash-code, we don't know the size of the target object until the forwarding phase. That is because we need to know whether or not an object moves at all, and we don't know that until we do the actual forwarding. It might be a possibility to make the pessimistic assumption that all objects would be moved, and thus might need to grow (if they have an i-hash and no gaps to store it), and then deal with the waste that gets produced, but some of the calculations about the split-points and the locations where objects would cross region boundaries would be thrown off still.

The new approach is much more flexible, and the implementation is much simpler (but it tends to produce more waste).

zhengyu123:
> > Can you articulate why object cannot grow?
> 
> Yes. The way PSParallelCompact currently works: it collects liveness information during the marking phase. At the end of it, we know precisely how many words are live in each reagion. In this phase, we also determine where objects cross region boundaries. Then, during the summary phase, it uses this information to determine which source regions get compacted into which target regions. It also determines the addresses at which a region gets split - that is the part below the split-point gets compacted into one region, and the part above the split point gets compacted into another region. Then, during the forwarding phase, all those infos are used to calculate the target addresses for each object in parallel. It is now possible to divide the forwarding work across several GC worker threads because we know exactly where each region gets compacted into.
> 

You can count liveness with/without object expansion during mark  phase, can you? given it reads object header and size anyway. I assume that only regions "overflow" due to expansion, can cause problem, but chance is very slim, special case them won't be so bad, right?



> The problem is that, with compact identity hash-code, we don't know the size of the target object until the forwarding phase. That is because we need to know whether or not an object moves at all, and we don't know that until we do the actual forwarding. It might be a possibility to make the pessimistic assumption that all objects would be moved, and thus might need to grow (if they have an i-hash and no gaps to store it), and then deal with the waste that gets produced, but some of the calculations about the split-points and the locations where objects would cross region boundaries would be thrown off still.
> 
> The new approach is much more flexible, and the implementation is much simpler (but it tends to produce more waste).

rkennke:
> > > Can you articulate why object cannot grow?
> > 
> > 
> > Yes. The way PSParallelCompact currently works: it collects liveness information during the marking phase. At the end of it, we know precisely how many words are live in each reagion. In this phase, we also determine where objects cross region boundaries. Then, during the summary phase, it uses this information to determine which source regions get compacted into which target regions. It also determines the addresses at which a region gets split - that is the part below the split-point gets compacted into one region, and the part above the split point gets compacted into another region. Then, during the forwarding phase, all those infos are used to calculate the target addresses for each object in parallel. It is now possible to divide the forwarding work across several GC worker threads because we know exactly where each region gets compacted into.
> 
> You can count liveness with/without object expansion during mark phase, can you?

Yes, I can. I actually implemented that in an earlier attemp.

> I assume that only regions "overflow" due to expansion, can cause problem, but chance is very slim, special case them won't be so bad, right?

The problem is not potential overflow. In-fact, overflow can not really happen. The problem is that all those calculations are assumed and required to be precise. Turning them into guesses breaks the algorithm.

tstuefe:
How does this algorithm deal with objects larger than region size?

rkennke:
> How does this algorithm deal with objects larger than region size?

Currently not particularily well: it doesn't move them at all (because they don't fit), and it *also* doesn't move other objects past them. There's a TODO to make it possible to move objects around large objects (don't have to be > region-sized, also something like 1/2 region sized objects may have difficulty to move). I'll address that in a follow-up PR.

zhengyu123:
> > How does this algorithm deal with objects larger than region size?
> 
> Currently not particularily well: it doesn't move them at all (because they don't fit), and it _also_ doesn't move other objects past them. There's a TODO to make it possible to move objects around large objects (don't have to be > region-sized, also something like 1/2 region sized objects may have difficulty to move). I'll address that in a follow-up PR.

Can you explain why has problem to move objects > 1/2 region size? Thanks.

rkennke:
> > > How does this algorithm deal with objects larger than region size?
> > 
> > 
> > Currently not particularily well: it doesn't move them at all (because they don't fit), and it _also_ doesn't move other objects past them. There's a TODO to make it possible to move objects around large objects (don't have to be > region-sized, also something like 1/2 region sized objects may have difficulty to move). I'll address that in a follow-up PR.
> 
> Are you talking about parallel specific? Can you explain why region size > object size > 1/2 region size scenario? Thanks.

I only used 1/2 region size as an example. It can affect any object: if there is not enough space in the target region, and there is no other preceding target region in the queue, we might not be able to move the object, or only move the object within its region. It's only that the problem becomes worse the larger the object is, and for objects > region-size it's currently guaranteed that we can not move it. Being able to move objects around large/un-movable objects would help a lot to reduce gaps.

tstuefe (approved):
Okay

PR #203, opened 2025-09-26
8368755: [Lilliput] Fix CompressedClassSpaceSize handling

rkennke:
We currently set the default CompressedClassSpaceSize to 128M - much lower than the current upstream default of 1G, and still much lower than the possible encoding range with 4-byte-headers of 512M. I think I did this earlier to make some tests pass and to address a problem with CDS which also wants a smallish share of it, but I have in-fact only hidden a bug.

The way it should work is to set the CompressedClassSpaceSize to 512M and let CDS take its share and use the rest for class-space. This is infact implemented by [JDK-8332514](https://bugs.openjdk.org/browse/JDK-8332514) since a while. However, there is a bug there that prevents it from working with 4-byte-headers: it has 4G encoding range hard-coded.

This change fixes:
- It reverts the default CCSS to 1G
- It sets CCSS to 512M when running with COH
- It makes the auto-sizing work by setting the correct encoding range
- It fixes a test that tries to use CCSS*2. I changed it to use CCSS/2, I think this is still in the spirit of the test ('use a different CCSS and see if it still works')

Testing:
- [x] tier1
- [x] tier2
- [x] gc/arguments/TestUseCompressedOopsErgoTools.java (affected by the change)
- [x] runtime/cds/appcds/dynamicArchive/DynamicLotsOfClasses.java (failed before, fixed by this change)
- [x] runtime/cds/appcds/LotsOfJRTClasses.java (failed before, fixed by this change)

### DIFF
--- a/src/hotspot/share/runtime/arguments.cpp
+++ b/src/hotspot/share/runtime/arguments.cpp
@@ -3768,6 +3768,9 @@ void Arguments::set_compact_headers_flags() {
   if (UseCompactObjectHeaders && FLAG_IS_DEFAULT(hashCode)) {
     hashCode = 6;
   }
+  if (UseCompactObjectHeaders && FLAG_IS_DEFAULT(CompressedClassSpaceSize)) {
+    FLAG_SET_DEFAULT(CompressedClassSpaceSize, 512 * M);
+  }
 #endif
 }
 
--- a/src/hotspot/share/cds/metaspaceShared.cpp
+++ b/src/hotspot/share/cds/metaspaceShared.cpp
@@ -1758,8 +1758,9 @@ char* MetaspaceShared::reserve_address_space_for_archives(FileMapInfo* static_ma
   const size_t ccs_begin_offset = align_up(archive_space_size, class_space_alignment);
   const size_t gap_size = ccs_begin_offset - archive_space_size;
 
-  // Reduce class space size if it would not fit into the Klass encoding range
-  constexpr size_t max_encoding_range_size = 4 * G;
+  // Reduce class space size if it would not fit into the maximum possible Klass encoding range. That
+  // range is defined by the narrowKlass size.
+  const size_t max_encoding_range_size = CompressedKlassPointers::max_klass_range_size();
   guarantee(archive_space_size < max_encoding_range_size - class_space_alignment, "Archive too large");
   if ((archive_space_size + gap_size + class_space_size) > max_encoding_range_size) {
     class_space_size = align_down(max_encoding_range_size - archive_space_size - gap_size, class_space_alignment);
--- a/src/hotspot/share/runtime/globals.hpp
+++ b/src/hotspot/share/runtime/globals.hpp
@@ -1401,7 +1401,7 @@ const int ObjectAlignmentInBytes = 8;
           "Maximum size of Metaspaces (in bytes)")                          \
           constraint(MaxMetaspaceSizeConstraintFunc,AfterErgo)              \
                                                                             \
-  product(size_t, CompressedClassSpaceSize, 128*M,                          \
+  product(size_t, CompressedClassSpaceSize, 1*G,                            \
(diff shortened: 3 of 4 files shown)


src/hotspot/share/runtime/arguments.cpp
```diff
@@ -3768,6 +3768,9 @@ void Arguments::set_compact_headers_flags() {
   if (UseCompactObjectHeaders && FLAG_IS_DEFAULT(hashCode)) {
     hashCode = 6;
   }
+  if (UseCompactObjectHeaders && FLAG_IS_DEFAULT(CompressedClassSpaceSize)) {
+    FLAG_SET_DEFAULT(CompressedClassSpaceSize, 512 * M);
+  }
```
tstuefe, line 3773: Still a bit ugly, since again hardcoded.

What I would do is:

1) move the call to `CompressedKlassPointers::pre_initialize()` at line 3789:

```
  if (UseCompressedClassPointers) {
    CompressedKlassPointers::pre_initialize();
  }
```

down into ``  from apply_ergo (just below) to before you set the CCSS.

Then, set the CCS with FLAG_SET_DEFAULT(CompressedClassSpaceSize, 512 * M);

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -3768,6 +3768,9 @@ void Arguments::set_compact_headers_flags() {
   if (UseCompactObjectHeaders && FLAG_IS_DEFAULT(hashCode)) {
     hashCode = 6;
   }
+  if (UseCompactObjectHeaders && FLAG_IS_DEFAULT(CompressedClassSpaceSize)) {
+    FLAG_SET_DEFAULT(CompressedClassSpaceSize, 512 * M);
```
tstuefe, line 3772: This may be a bit less ugly and hard-coded:

```suggestion
    FLAG_SET_DEFAULT(CompressedClassSpaceSize, CompressedKlassPointers::max_klass_range_size_coh);
```

with this patch in addition:

```
diff --git a/src/hotspot/share/oops/compressedKlass.cpp b/src/hotspot/share/oops/compressedKlass.cpp
index d7c97d3c8d5..bf45f4ecd78 100644
--- a/src/hotspot/share/oops/compressedKlass.cpp
+++ b/src/hotspot/share/oops/compressedKlass.cpp
@@ -47,6 +47,7 @@ size_t CompressedKlassPointers::_protection_zone_size = 0;
 size_t CompressedKlassPointers::max_klass_range_size() {
 #ifdef _LP64
   const size_t encoding_allows = nth_bit(narrow_klass_pointer_bits() + max_shift());
+  assert(!UseCompactObjectHeaders || max_klass_range_size_coh == encoding_allows, "Sanity");
   constexpr size_t cap = 4 * G;
   return MIN2(encoding_allows, cap);
 #else
diff --git a/src/hotspot/share/oops/compressedKlass.hpp b/src/hotspot/share/oops/compressedKlass.hpp
index 64b9fcf9c82..4f94675d8ed 100644
--- a/src/hotspot/share/oops/compressedKlass.hpp
+++ b/src/hotspot/share/oops/compressedKlass.hpp
@@ -192,6 +192,10 @@ class CompressedKlassPointers : public AllStatic {
   // resulting from the current encoding settings (base, shift), capped to a certain max. value.
   static size_t max_klass_range_size();
 
+  // For use before pre-initialization
+  static constexpr size_t max_klass_range_size_coh =
+      nth_bit(narrow_klass_pointer_bits_coh + max_shift_coh);
+
   // On 64-bit, we need the class space to confine Klass structures to the encoding range, which is determined
   // by bit size of narrowKlass IDs and the shift. On 32-bit, we support compressed class pointer only
   // "pro-forma": narrowKlass have the same size as addresses (32 bits), and therefore the encoding range is
```

rkennke:
GHA failures look like unrelated infra issues.

rkennke:
Thank you!

/integrate

PR #207, opened 2026-03-02
8372151: Lilliput2: Don't copy mark-word in Object.clone()

rkennke:
We've observed heap corruption with Lilliput2 (see bug report). The root cause is that when we do an Object.clone(), we also copy the mark-word to the new object, which is wrong: we might copy an active lock-state, which is wrong, and we might also copy the hash bits, which results in wrong size calculations in the new object.

The fast C2 copy routine only copies 8-byte-words, and I did not want to mess with that. I'm emitting a separate load&store to copy the first 4 bytes, and only use the 8-bytes-copy-routine on the remaining 8-byte words, and avoid copying the mark-word altogether.

For the runtime case, we copy the full object, and re-initialize the mark word with its prototype header afterwards. (init_mark() preserves the hash-bits, and is used by full GCs for initializing the mark-words of object copies.)

Testing:
 - [x] The problem could reliably be reproduced by running: `java -XX:+UseCompactObjectHeaders -XX:+VerifyDuringGC -Xmx9g -Xms9g -jar renaissance-gpl-0.16.1.jar dotty` and does not fail anymore with the fix
 - [x] tier1 +UCOH
 - [x] tier1 -UCOH

---------
### Progress

### Issue
 * [JDK-8372151](https://bugs.openjdk.org/browse/JDK-8372151): Lilliput2: Don't copy mark-word in Object.clone() (**Bug** - P4)

### Reviewers
 * [Y. Srinivas Ramakrishna](https://openjdk.org/census#ysr) (@ysramakrishna - **Reviewer**)
 * [Oli Gillespie](https://openjdk.org/census#ogillespie) (@olivergillespie - Author) ⚠️ Review applies to [dab4136b](https://git.openjdk.org/lilliput/pull/207/files/dab4136b3f2622ba2aba5f20ea1fbaf127f31d1d)
 * [Paul Hohensee](https://openjdk.org/census#phh) (@phohensee - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/207/head:pull/207` \
`$ git checkout pull/207`

`$ git checkout pull/207` \
`$ git pull https://git.openjdk.org/lilliput.git pull/207/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 207`

`$ git pr show -t 207`

</details>

<a href="https://git.openjdk.org/lilliput/pull/207.diff">https://git.openjdk.org/lilliput/pull/207.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/207#issuecomment-3984773162)
</details>

### DIFF
--- a/src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
+++ b/src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
@@ -714,9 +714,7 @@ int BarrierSetC2::arraycopy_payload_base_offset(bool is_array) {
       // Exclude length to copy by 8 bytes words.
       base_off += sizeof(int);
     } else {
-      if (UseCompactObjectHeaders) {
-        base_off = 0; /* FIXME */
-      } else {
+      if (!UseCompactObjectHeaders) {
         // Include klass to copy by 8 bytes words.
         base_off = instanceOopDesc::klass_offset_in_bytes();
       }
@@ -728,6 +726,39 @@ int BarrierSetC2::arraycopy_payload_base_offset(bool is_array) {
 
 void BarrierSetC2::clone(GraphKit* kit, Node* src_base, Node* dst_base, Node* size, bool is_array) const {
   int base_off = arraycopy_payload_base_offset(is_array);
+  if (UseCompactObjectHeaders && !is_aligned(base_off, BytesPerLong) &&
+      !kit->gvn().type(src_base)->isa_aryptr()) {
+    guarantee(is_aligned(base_off, BytesPerInt), "must be 4-bytes aligned");
+    // The optimized copy routine only copies 8-byte words. For this reason, we must
+    // copy the 4 bytes at offset 4 separately.
+    // Use the correct field-specific alias derived from the typed address, matching
+    // the pattern in PhaseMacroExpand::generate_arraycopy (macroArrayCopy.cpp).
+    // Using AliasIdxRaw would create a mismatch between the typed address and the
+    // raw memory chain, causing an escape analysis assertion failure.
+    //
+    // Skip this when src_base has an array type. With StressReflectiveCode, the
+    // instance path of the clone can be live in the IR even when the type system
+    // knows src_base is an array. The pre-copy is unnecessary on such paths (they
+    // are unreachable at runtime), and creating a LoadNode at the array length
+    // offset would assert (LoadRangeNode required).
+    Node* sptr = kit->basic_plus_adr(src_base, base_off);
+    Node* dptr = kit->basic_plus_adr(dst_base, base_off);
+    const TypePtr* s_adr_type = kit->gvn().type(sptr)->is_ptr();
+    const TypePtr* d_adr_type = kit->gvn().type(dptr)->is_ptr();
+    uint s_alias_idx = Compile::current()->get_alias_index(s_adr_type);
+    uint d_alias_idx = Compile::current()->get_alias_index(d_adr_type);
+    Node* first = kit->gvn().transform(LoadNode::make(kit->gvn(), kit->control(), kit->memory(s_alias_idx),
+                                       sptr, s_adr_type, TypeInt::INT, T_INT,
+                                       MemNode::unordered));
+    Node* st = kit->gvn().transform(StoreNode::make(kit->gvn(), kit->control(), kit->memory(d_alias_idx),
+                                                    dptr, d_adr_type,
+                                                    first, T_INT, MemNode::unordered));
+    kit->set_memory(st, d_alias_idx);
+    kit->record_for_igvn(st);
+    base_off += sizeof(jint);
+    guarantee(is_aligned(base_off, BytesPerLong), "must be 8-bytes aligned");
+  }
+
   Node* payload_size = size;
   Node* offset = kit->MakeConX(base_off);
   payload_size = kit->gvn().transform(new SubXNode(payload_size, offset));
--- a/src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
+++ b/src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
@@ -882,7 +882,10 @@ void BarrierSetC2::clone_in_runtime(PhaseMacroExpand* phase, ArrayCopyNode* ac,
 
   // The native clone we are calling here expects the object size in words.
   // Add header/offset size to payload size to get object size.
-  Node* const base_offset = phase->MakeConX(arraycopy_payload_base_offset(ac->is_clone_array()) >> LogBytesPerLong);
+  // Use the actual offset stored in the ArrayCopyNode (in bytes), not
+  // arraycopy_payload_base_offset(), because clone() may have bumped the
+  // offset past a 4-byte pre-copy for compact object headers.
+  Node* const base_offset = phase->transform_later(new URShiftXNode(ac->in(ArrayCopyNode::SrcPos), phase->intcon(LogBytesPerLong)));
   Node* const full_size = phase->transform_later(new AddXNode(size, base_offset));
   // HeapAccess<>::clone expects size in heap words.
   // For 64-bits platforms, this is a no-operation.
--- a/src/hotspot/share/oops/accessBackend.inline.hpp
+++ b/src/hotspot/share/oops/accessBackend.inline.hpp
@@ -327,7 +327,7 @@ inline void RawAccessBarrier<decorators>::clone(oop src, oop dst, size_t size) {
                                             reinterpret_cast<jlong*>((oopDesc*)dst),
                                             align_object_size(size) / HeapWordsPerLong);
   // Clear the header
-  dst->init_mark();
+  dst->set_mark(dst->prototype_mark());
 }
 
 #endif // SHARE_OOPS_ACCESSBACKEND_INLINE_HPP
--- a/src/hotspot/share/oops/accessBackend.inline.hpp
+++ b/src/hotspot/share/oops/accessBackend.inline.hpp
@@ -327,7 +327,7 @@ inline void RawAccessBarrier<decorators>::clone(oop src, oop dst, size_t size) {
                                             reinterpret_cast<jlong*>((oopDesc*)dst),
                                             align_object_size(size) / HeapWordsPerLong);
   // Clear the header
-  dst->set_mark(dst->prototype_mark());
+  dst->init_mark();
 }
 
 #endif // SHARE_OOPS_ACCESSBACKEND_INLINE_HPP
--- a/src/hotspot/share/gc/g1/g1FullGCCompactTask.cpp
+++ b/src/hotspot/share/gc/g1/g1FullGCCompactTask.cpp
@@ -63,7 +63,7 @@ void G1FullGCCompactTask::copy_object_to_new_location(oop obj) {
   Copy::aligned_conjoint_words(obj_addr, destination, size);
 
   // There is no need to transform stack chunks - marking already did that.
-  cast_to_oop(destination)->init_mark();
+  cast_to_oop(destination)->reinit_mark();
   cast_to_oop(destination)->initialize_hash_if_necessary(obj);
   assert(cast_to_oop(destination)->klass() != nullptr, "should have a class");
 }
--- a/src/hotspot/share/gc/parallel/psParallelCompact.cpp
+++ b/src/hotspot/share/gc/parallel/psParallelCompact.cpp
@@ -2374,7 +2374,7 @@ void MoveAndUpdateClosure::do_addr(HeapWord* addr, size_t words) {
     assert(FullGCForwarding::is_forwarded(cast_to_oop(source())), "inv");
     assert(FullGCForwarding::forwardee(cast_to_oop(source())) == cast_to_oop(destination()), "inv");
     Copy::aligned_conjoint_words(source(), copy_destination(), words);
-    cast_to_oop(copy_destination())->init_mark();
+    cast_to_oop(copy_destination())->reinit_mark();
   }
 
   update_state(words);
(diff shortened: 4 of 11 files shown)


src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
```diff
@@ -728,6 +726,39 @@ int BarrierSetC2::arraycopy_payload_base_offset(bool is_array) {
 
 void BarrierSetC2::clone(GraphKit* kit, Node* src_base, Node* dst_base, Node* size, bool is_array) const {
   int base_off = arraycopy_payload_base_offset(is_array);
+  if (UseCompactObjectHeaders && !is_aligned(base_off, BytesPerLong) &&
+      !kit->gvn().type(src_base)->isa_aryptr()) {
```
JohnTortugo: NIT: is this same as `!is_array` ?

JohnTortugo (commented):
I'll submit the patch to our internal CI for testing.

src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
rkennke: No. I thought so too, but then got an assert elsewhere that I should not load the array-length using a LoadInt. The way that is_array is determined is different from type(src_base)->isa_aryptr(), it uses the Klass of the object, and at that point it might not yet figured out that this is an array and still treat it as a Klass. (I wonder if that could be a problem for other code that depends on is_array?)

src/hotspot/share/oops/accessBackend.inline.hpp
```diff
@@ -327,7 +327,7 @@ inline void RawAccessBarrier<decorators>::clone(oop src, oop dst, size_t size) {
                                             reinterpret_cast<jlong*>((oopDesc*)dst),
                                             align_object_size(size) / HeapWordsPerLong);
   // Clear the header
-  dst->init_mark();
+  dst->set_mark(dst->prototype_mark());
```
ysramakrishna: Have we looked at other uses of `init_mark()` to make sure they don't suffer from this issue? I do see that the spec comment for `oopDesc::init_mark()` says the right thing in terms of `re-initializing the mark word` but a clearer statement on where one vs the other should be used would help with on-going maintenance of this code.

I also wonder if there is a "consistency check" that may be possible to match the size of an object post-initialization vs the space allocated for it. That might serve to further steelman this code, or at least make the detection of these errors closer to allocation?

ysramakrishna (commented):
Flushing an old comment; will take a fresh look after looking more closely at the test failures that Cesar uncovered.

olivergillespie (approved):
Makes sense to me (but I don't know C2 well enough to check that part).

ysramakrishna (approved):
Re-tagging @JohnTortugo (and alternately @phohensee ) to please (re)review the C2 changes, please.

phohensee (approved):

rkennke:
It's not easy to trigger the bug. I managed to do it with the following modifications:

1. New RefPayload class (lines 191-212): Has only Object reference fields (no primitives) so HotSpot places a narrowOop at offset 4 - where the clone pre-copy does a raw T_INT load/store. When mixing int and ref fields, the field layouter would place the int fields first.
2. cloneRefPayload() helper (lines 230-234): Allocates the source object inline so C2 can constant-fold the LoadN from the clone expansion. This reduces the pre-copy StoreI's outcnt to 1, enabling IGVN's StoreNode::Ideal to walk the memory chain and hit the StoreN-vs-StoreI assertion.
3. testCloneRefPayload() method (lines 249-257): Tight loop calling cloneRefPayload 100K times to trigger C2 compilation.
4. Argument-based dispatch (lines 236-244): main dispatches between testClonePayload() (GC stress, no args) and testCloneRefPayload() (C2 regression, clone-ref arg) to keep the GC tests isolated from the C2 assertion.
5. Three new C2 test configurations (C2-Serial, C2-Parallel, C2-G1): Pass clone-ref argument and use -XX:-TieredCompilation to force C2.

You can find the modified test here: https://github.com/openjdk/lilliput/blob/daea5edef6eaf56bef1ac157fa872d7f81896c27/test/hotspot/jtreg/gc/stress/ihash/TestStressIHash.java

rkennke:
I filed a new bug about the C2 issue: https://bugs.openjdk.org/browse/JDK-8381489

olivergillespie:
I'm investigating a possible bug in this implementation with Shenandoah. My limited understanding - this change splits the copy into two, a 4-byte copy and then the original ArrayCopyNode for the remainder. Shenandoah's [clone_at_expansion](https://github.com/openjdk/lilliput/blob/daea5edef6eaf56bef1ac157fa872d7f81896c27/src/hotspot/share/gc/shenandoah/c2/shenandoahBarrierSetC2.cpp#L899) only applies to the ArrayCopyNode, so during forwarding we might copy an unresolved forwarding pointer.

I'm still investigating and will create an issue (and maybe a fix) once I have more info.

rkennke:
> I'm investigating a possible bug in this implementation with Shenandoah. My limited understanding - this change splits the copy into two, a 4-byte copy and then the original ArrayCopyNode for the remainder. Shenandoah's [clone_at_expansion](https://github.com/openjdk/lilliput/blob/daea5edef6eaf56bef1ac157fa872d7f81896c27/src/hotspot/share/gc/shenandoah/c2/shenandoahBarrierSetC2.cpp#L899) only applies to the ArrayCopyNode, so during forwarding we might copy an unresolved forwarding pointer.
> 
> I'm still investigating and will create an issue (and maybe a fix) once I have more info.

I don't think we would ever copy an unresolved fwd-pointer because the mutator thread never gets to see a from-space object.

Or do you mean that it might copy a pointer to a from-space object that lives at offset 4? That could be possible. I haven't checked this particular situation.

Do you have a specific failure that is related to this? Or are you investigating proactively?

/Roman

olivergillespie:
Yes, I'm seeing a specific crash, and it seems to bisect to this commit.

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f3df371766f, pid=65799, tid=65920
#
# JRE version: OpenJDK Runtime Environment (27.0) (build 27-internal-adhoc.ogillesp.lilliput)
# Java VM: OpenJDK 64-Bit Server VM (27-internal-adhoc.ogillesp.lilliput, mixed mode, tiered, compressed oops, compact obj headers, shenandoah gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0xfad66f]  void ShenandoahMark::mark_loop_work<ShenandoahMarkRefsClosure<(ShenandoahGenerationType)0>, (ShenandoahGenerationType)0, true, (StringDedupMode)0>(ShenandoahMarkRefsClosure<(ShenandoahGenerationType)0>*, unsigned short*, unsigned int, TaskTerminator*, StringDedup::Requests*) [clone .isra.0]+0x2bf
#
# No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# If you would like to submit a bug report, please visit:
#   https://bugreport.java.com/bugreport/crash.jsp
#

---------------  S U M M A R Y ------------

Command Line: -XX:+UseShenandoahGC -XX:ParallelGCThreads=8 -XX:ConcGCThreads=4 -XX:+UseCompactObjectHeaders -Xmx4g -Xms4g -XX:ErrorFile=/tmp/repro-results-commit-lilliput-non-gen/hs_err_run8_pid%p.log /tmp/shen-crash/dacapo-23.11-MR2-chopin.jar -s large -n 1 xalan

Host: Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz, 72 cores, 137G, Amazon Linux release 2 (Karoo)
Time: Fri Apr 10 16:00:02 2026 UTC elapsed time: 8.459867 seconds (0d 0h 0m 8s)

---------------  T H R E A D  ---------------

Current thread (0x00007f3dec0f2680):  WorkerThread "Shenandoah GC Threads#0" [id=65920, stack(0x00007f3df14ff000,0x00007f3df15ff000) (1024K)]

Stack: [0x00007f3df14ff000,0x00007f3df15ff000],  sp=0x00007f3df15fdae0,  free space=1018k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
V  [libjvm.so+0xfad66f]  void ShenandoahMark::mark_loop_work<ShenandoahMarkRefsClosure<(ShenandoahGenerationType)0>, (ShenandoahGenerationType)0, true, (StringDedupMode)0>(ShenandoahMarkRefsClosure<(ShenandoahGenerationType)0>*, unsigned short*, unsigned int, TaskTerminator*, StringDedup::Requests*) [clone .isra.0]+0x2bf  (shenandoahMark.inline.hpp:78)
V  [libjvm.so+0xfb3ac7]  ShenandoahMark::mark_loop(unsigned int, TaskTerminator*, ShenandoahGenerationType, bool, StringDedupMode, StringDedup::Requests*)+0x787  (shenandoahMark.cpp:74)
V  [libjvm.so+0xf5df04]  ShenandoahConcurrentMarkingTask<(ShenandoahGenerationType)0>::work(unsigned int)+0xb4  (shenandoahConcurrentMark.cpp:63)
V  [libjvm.so+0x121ff90]  WorkerThread::run()+0x80  (workerThread.cpp:71)
V  [libjvm.so+0x114ac68]  Thread::call_run()+0xa8  (thread.cpp:243)
V  [libjvm.so+0xe3554b]  thread_native_entry(Thread*)+0xcb  (os_linux.cpp:931)

siginfo: si_signo: 11 (SIGSEGV), si_code: 2 (SEGV_ACCERR), si_addr: 0x00000000b600000c

Registers:
RAX=0x00000000b6000000, RBX=0xfffe000000000000, RCX=0x000000000000000a, RDX=0x00007f3df3e20950
RSP=0x00007f3df15fdae0, RBP=0x00007f3df15fdbd0, RSI=0x00000007fd0d11b8, RDI=0x00007f3df15fdcc0
R8 =0x00007f3df3e5e170, R9 =0x00007f3d97ff0088, R10=0x0000000000000049, R11=0x00007f3df3568860
R12=0x00000007fff0f070, R13=0x00007f3df15fdb00, R14=0x00007f3dec0e67c0, R15=0x00007f3dec0e67c0
RIP=0x00007f3df371766f, EFLAGS=0x0000000000010206, CSGSFS=0x002b000000000033, ERR=0x0000000000000004
  TRAPNO=0x000000000000000e
```

> Or do you mean that it might copy a pointer to a from-space object that lives at offset 4? That could be possible. I haven't checked this particular situation.

I don't know exactly, still trying to understand. Ignoring my hypotheses, what I observe is that it always crashes trying to access compressed class space base + 0x0c. 

My reproducer is:

```
java \
        -XX:+UseShenandoahGC \
        -XX:+UnlockExperimentalVMOptions \
        -XX:ParallelGCThreads=8 -XX:ConcGCThreads=4 \
	-XX:+UseCompactObjectHeaders \
        -Xmx4g -Xms4g \
        -jar dacapo-23.11-MR2-chopin.jar -s large -n 1 xalan
``` 

It crashes a few times per 100 runs on average, for me.

rkennke:
I think you are right: BSC2::clone() generates the ArrayCopyNode plus the preceding 4-byte copy. ShBSC2::clone_at_expansion() only handles the expansion of the actual ArrayCopyNode, but not the preceding 4-byte-copy. However, that is necessary to resolve any references to from-space-objects that might have been copied to the new object.

Possible solutions:
- Don't break up the array-copy into the 4-byte plus ArrayCopyNode in BSC2::clone() but instead handle that in BSC2::clone_at_expansion(). For this we would have to pass the real start-offset in the ArrayCopyNode, and only align the copy routine in clone_at_expansion. This might require significant rework in how C2 and the BSC2 handles clone (not sure).
- Alternatively, override clone() in ShBSC2 (in addition to clone_at_expansion) and handle the preceding 4 bytes correctly. This avoids any tricky problems in shared C2/BSC2 code at the expense of more mess in ShBSC2.

olivergillespie:
Thanks.

What about doing the whole copy including mark-word like before, but then overwriting the mark-word with a fresh init_mark()? That gives us minimal extra barrier management, we just need to take care of writing the mark-word.

I'm playing with implementations but I'm starting from zero with C2.

rkennke:
> Thanks.
> 
> What about doing the whole copy including mark-word like before, but then overwriting the mark-word with a fresh init_mark()? That gives us minimal extra barrier management, we just need to take care of writing the mark-word.
> 
> I'm playing with implementations but I'm starting from zero with C2.

This could be done, however, you need the Klass of the object to do that. Which means: load and decode the Klass* from the original object or the copy, load the mark-prototype out of that Klass, store that in the copy, and all of that in C2 code. That is 2 extra loads and 1 extra store, which might be a performance concern on small objects.

olivergillespie:
Hmm. Very small instances (<=ArrayCopyLoadStoreMaxElem, default 8 = 64 bytes) are [replaced](https://github.com/openjdk/lilliput/blob/daea5edef6eaf56bef1ac157fa872d7f81896c27/src/hotspot/share/opto/arraycopynode.cpp#L558-L563) with individual loads and stores and don't go through clone_at_expansion, right? So the extra loads/stores would be specifically felt for small-ish objects I guess.

olivergillespie:
And could we make it 1 load and 1 store by loading from destination (which already has its header initialized, I think...)?

olivergillespie:
I've been trying to implement a good fix, but I'm struggling a lot with C2. I wrote up the bug here: https://bugs.openjdk.org/browse/JDK-8372151. And here's my best candidate fix so far: https://gist.github.com/olivergillespie/5905646f2d854784b3eb62ab4b60760e.

PR #213, opened 2026-04-10
8381999: [Lilliput2] Increase region size for Parallel Full GC

rudraroop:
Linked issue :  https://bugs.openjdk.org/browse/JDK-8381999

## Problem

Parallel Full GC (`psParallelCompactNew`) uses a fixed **0.5MB** region size. Normal (non-maximum) compaction can't move humongous objects ( objects bigger than region size) past other humongous objects. This can lead to a state where the GC becomes unproductive and immediately triggers maximum compaction. An example of this can be found in [JDK-8375467: [Lilliput] Unproductive Full GC with Parallel](https://bugs.openjdk.org/browse/JDK-8375467)

## Change

Replace the fixed region size with a dynamic calculation:
```
region_size = max_waste_percent × heap_size / num_workers
```

Each worker can leave at most one region of waste at the boundary between worker sets (the last region it processes can be almost completely empty). With _**N** workers_ and _**R** regions per worker_, total boundary waste is at most N regions out of N * R total, i.e. 1/R of the heap.

image

Defining 5% as the maximum tolerable boundary waste gives Regions per worker ≥ 20. So our calculation is based on each worker having 20 regions to work with. From there, the region size follows directly:
```
total_regions = max_workers * 20
region_size   = heap_size / total_regions
              = heap_size / (max_workers * 20)
```

The result is rounded up to the nearest power of 2 (for alignment), and floored at the original 0.5MB default to avoid unreasonably small regions on tiny heaps.

For a 256MB heap with 4 workers this gives ~4MB regions; for a 4GB heap with 8 workers, ~32MB. Objects that were humongous at 0.5MB (like the 1MB arrays in the JDK-8375467 reproducer) now fit comfortably within a single region and compact normally.

## Regressions?

Each worker gets fewer regions (each being larger). Objects still slide freely across the same worker's regions, so compaction for each worker is unchanged. Bin-packing waste at region tops actually decreases slightly with fewer, larger regions. Only the addresses where worker boundaries fall is changed.

---------
### Progress

### Issue
 * [JDK-8381999](https://bugs.openjdk.org/browse/JDK-8381999): [Lilliput2] Increase region size for Parallel Full GC (**Enhancement** - P4)

### Reviewers
 * [Roman Kennke](https://openjdk.org/census#rkennke) (@rkennke - **Reviewer**) ⚠️ Review applies to [3a53afe9](https://git.openjdk.org/lilliput/pull/213/files/3a53afe9912700ae1b6d736e4e6538aab30b8751)
 * [Oli Gillespie](https://openjdk.org/census#ogillespie) (@olivergillespie - Author)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/213/head:pull/213` \
`$ git checkout pull/213`

`$ git checkout pull/213` \
`$ git pull https://git.openjdk.org/lilliput.git pull/213/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 213`

`$ git pr show -t 213`

</details>

<a href="https://git.openjdk.org/lilliput/pull/213.diff">https://git.openjdk.org/lilliput/pull/213.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/213#issuecomment-4224057980)
</details>

### DIFF
--- a/src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
+++ b/src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
@@ -375,7 +375,7 @@ void PSParallelCompactNew::post_compact()
 
 void PSParallelCompactNew::setup_regions_parallel() {
   static const size_t REGION_SIZE_WORDS = compute_region_size();
-  log_debug(gc, compaction)("RegionSizeBytes: %zu bytes (%.2f MB)", REGION_SIZE_WORDS * HeapWordSize, (double)(REGION_SIZE_WORDS * HeapWordSize) / (1024.0 * 1024.0));
+  log_debug(gc, compaction)("Region size: %zu bytes (%.2f MB)", REGION_SIZE_WORDS * HeapWordSize, (double)(REGION_SIZE_WORDS * HeapWordSize) / (1024.0 * 1024.0));
 
   size_t num_regions = 0;
   for (uint i = old_space_id; i < last_space_id; ++i) {
@@ -423,7 +423,7 @@ void PSParallelCompactNew::setup_regions_parallel() {
 }
 
 size_t PSParallelCompactNew::compute_region_size() {
-  // Default 0.5MB Region Size
+  // Minimum 0.5MB Region Size
   const size_t FLOOR_REGION_SIZE_WORDS = (SpaceAlignment / HeapWordSize);
   size_t total_heap_words = 0;
   for (uint i = old_space_id; i < last_space_id; ++i) {
@@ -431,8 +431,7 @@ size_t PSParallelCompactNew::compute_region_size() {
   }
 
   // Per-worker region count for dynamic region sizing
-  const uint MaxWastePercent = 5;  // 5% waste threshold
-  const uint RegionsPerWorker = 100 / MaxWastePercent;
+  const uint RegionsPerWorker = 20;  // Based on 5% boundary waste threshold
   const uint max_workers = ParallelScavengeHeap::heap()->workers().max_workers();
   const size_t total_regions_count = (size_t)max_workers * RegionsPerWorker;
   const size_t DYNAMIC_REGION_SIZE_WORDS = round_up_power_of_2(total_heap_words / total_regions_count);
--- a/src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
+++ b/src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
@@ -316,7 +316,7 @@ void PSParallelCompactNew::post_compact()
         }
       }
     }
-    log_develop_debug(gc, compaction)("total live: %zu, total waste: %zu, ratio: %f", total_live, total_waste, ((float)total_waste)/((float)(total_live + total_waste)));
+    log_debug(gc, compaction)("total live: %zu, total waste: %zu, ratio: %f", total_live, total_waste, ((float)total_waste)/((float)(total_live + total_waste)));
   }
   {
     FREE_C_HEAP_ARRAY(PCRegionData*, _per_worker_region_data);
@@ -374,7 +374,10 @@ void PSParallelCompactNew::post_compact()
 }
 
 void PSParallelCompactNew::setup_regions_parallel() {
-  static const size_t REGION_SIZE_WORDS = (SpaceAlignment / HeapWordSize);
+
+  static const size_t REGION_SIZE_WORDS = compute_region_size();
+  log_debug(gc, compaction)("RegionSizeBytes: %zu bytes (%.2f MB)", REGION_SIZE_WORDS * HeapWordSize, (double)(REGION_SIZE_WORDS * HeapWordSize) / (1024.0 * 1024.0));
+
   size_t num_regions = 0;
   for (uint i = old_space_id; i < last_space_id; ++i) {
     MutableSpace* const space = _space_info[i].space();
@@ -418,6 +421,27 @@ void PSParallelCompactNew::setup_regions_parallel() {
   }
   _num_regions = region_idx;
   log_info(gc)("Number of regions: %zu", _num_regions);
+}
+
+size_t PSParallelCompactNew::compute_region_size() {
+  
+  // Default 0.5MB Region Size
+  const size_t FLOOR_REGION_SIZE_WORDS = (SpaceAlignment / HeapWordSize);
+  
+  size_t total_heap_words = 0;
+  for (uint i = old_space_id; i < last_space_id; ++i) {
+    total_heap_words += _space_info[i].space()->capacity_in_words();
+  }
+
+  // Per-worker region count for dynamic region sizing
+  const uint MaxWastePercent = 5;  // 5% waste threshold
+  const uint RegionsPerWorker = 100 / MaxWastePercent;
+  const uint max_workers = ParallelScavengeHeap::heap()->workers().max_workers();
+  const size_t total_regions_count = (size_t)max_workers * RegionsPerWorker;
+  const size_t DYNAMIC_REGION_SIZE_WORDS = round_up_power_of_2(total_heap_words / total_regions_count);
+  
+  return MAX2(DYNAMIC_REGION_SIZE_WORDS, FLOOR_REGION_SIZE_WORDS);
+
 }
 
 void PSParallelCompactNew::setup_regions_serial() {
--- a/src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
+++ b/src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
@@ -423,18 +423,18 @@ void PSParallelCompactNew::setup_regions_parallel() {
 
 size_t PSParallelCompactNew::calculate_region_size() {
   // Minimum 0.5MB region size
-  const size_t floor_region_size_words = (SpaceAlignment / HeapWordSize);
+  static const size_t FLOOR_REGION_SIZE_WORDS = (SpaceAlignment / HeapWordSize);
   size_t total_heap_words = 0;
   for (uint i = old_space_id; i < last_space_id; ++i) {
     total_heap_words += _space_info[i].space()->capacity_in_words();
   }
 
   // Per-worker region count for dynamic region sizing
-  const uint regions_per_worker = 20;  // Based on 5% boundary waste threshold
+  static const uint REGIONS_PER_WORKER = 20;  // Based on 5% boundary waste threshold
   const uint max_workers = ParallelScavengeHeap::heap()->workers().max_workers();
-  const size_t total_regions_count = (size_t)max_workers * regions_per_worker;
+  const size_t total_regions_count = (size_t)max_workers * REGIONS_PER_WORKER;
   const size_t dynamic_region_size_words = total_heap_words / total_regions_count;
-  return round_up_power_of_2(MAX2(dynamic_region_size_words, floor_region_size_words));
+  return round_up_power_of_2(MAX2(dynamic_region_size_words, FLOOR_REGION_SIZE_WORDS));
 }
 
 void PSParallelCompactNew::setup_regions_serial() {
--- a/src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
+++ b/src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
@@ -374,7 +374,7 @@ void PSParallelCompactNew::post_compact()
 }
 
 void PSParallelCompactNew::setup_regions_parallel() {
-  static const size_t REGION_SIZE_WORDS = compute_region_size();
+  const size_t REGION_SIZE_WORDS = compute_region_size();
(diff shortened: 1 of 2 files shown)


src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
```diff
@@ -420,6 +423,27 @@ void PSParallelCompactNew::setup_regions_parallel() {
...
+  const size_t FLOOR_REGION_SIZE_WORDS = (SpaceAlignment / HeapWordSize);
+  
+  size_t total_heap_words = 0;
+  for (uint i = old_space_id; i < last_space_id; ++i) {
+    total_heap_words += _space_info[i].space()->capacity_in_words();
+  }
+
+  // Per-worker region count for dynamic region sizing
+  const uint MaxWastePercent = 5;  // 5% waste threshold
+  const uint RegionsPerWorker = 100 / MaxWastePercent;
```
olivergillespie: I think it's fine to just say regions_per_worker = 20. The waste thing is kinda interesting but it's not a true overall limit on waste anyway.

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
```diff
@@ -420,6 +422,23 @@ void PSParallelCompactNew::setup_regions_parallel() {
   log_info(gc)("Number of regions: %zu", _num_regions);
 }
 
+size_t PSParallelCompactNew::compute_region_size() {
+  // Default 0.5MB Region Size
```
olivergillespie: Not really default, just minimum. And that's kinda obvious from the code.

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
```diff
@@ -420,6 +422,23 @@ void PSParallelCompactNew::setup_regions_parallel() {
...
+  const size_t FLOOR_REGION_SIZE_WORDS = (SpaceAlignment / HeapWordSize);
+  size_t total_heap_words = 0;
+  for (uint i = old_space_id; i < last_space_id; ++i) {
+    total_heap_words += _space_info[i].space()->capacity_in_words();
```
olivergillespie, line 429: How does this work with a dynamic heap size?

Say -Xms100M -Xmx100G. This will be set on first run - will sum(capacity) be 100M at that point? But then later on if we expand the heap, we'll still be using the old values?

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
```diff
@@ -374,7 +374,9 @@ void PSParallelCompactNew::post_compact()
 }
 
 void PSParallelCompactNew::setup_regions_parallel() {
-  static const size_t REGION_SIZE_WORDS = (SpaceAlignment / HeapWordSize);
+  static const size_t REGION_SIZE_WORDS = compute_region_size();
+  log_debug(gc, compaction)("RegionSizeBytes: %zu bytes (%.2f MB)", REGION_SIZE_WORDS * HeapWordSize, (double)(REGION_SIZE_WORDS * HeapWordSize) / (1024.0 * 1024.0));
```
olivergillespie: `RegionSizeBytes: 123 bytes` is a bit clumsy. Perhaps `Region size: 123 bytes (0.123MB)`.

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
rudraroop: Makes sense - changed the wording

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
rudraroop: Done

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
rudraroop: Yeah, removed the MaxWastePercent step from the calculation and just kept `RegionsPerWorker = 20` but kept the comment about 5% waste just for clarity

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
rudraroop: I think this is the main problem with keeping REGION_SIZE_WORDS as a `static const` in the `setup_regions_parallel` function. 

In my understanding, this means REGION_SIZE_WORDS will only be computed using the compute_region_size function on the first run, and fail to adapt to changing heap size on subsequent runs

Dropping `static` from the REGION_SIZE_WORDS definition should fix this?

olivergillespie: I didn't actually check if the capacity_in_words changes as the heap shrinks/grows but I guess it must. I think recalculating makes sense, it should be a trivial overhead. It does mean the definition of humongous changes over time, which I don't love, but I don't see a clearly better option.

olivergillespie:
I agree with increasing the region size and I think the chosen formula is reasonable.

Style-wise you've got quite a range of variable name casing. Please follow https://wiki.openjdk.org/spaces/HotSpot/pages/11829275/StyleGuide#StyleGuide-Names.

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
```diff
@@ -374,7 +374,9 @@ void PSParallelCompactNew::post_compact()
 }
 
 void PSParallelCompactNew::setup_regions_parallel() {
-  static const size_t REGION_SIZE_WORDS = (SpaceAlignment / HeapWordSize);
+  const size_t REGION_SIZE_WORDS = compute_region_size();
+  log_debug(gc, compaction)("Region size: %zu bytes (%.2f MB)", REGION_SIZE_WORDS * HeapWordSize, (double)(REGION_SIZE_WORDS * HeapWordSize) / (1024.0 * 1024.0));
```
olivergillespie: You can use [EXACTFMT](https://github.com/openjdk/lilliput/blob/daea5edef6eaf56bef1ac157fa872d7f81896c27/src/hotspot/share/utilities/globalDefinitions.hpp#L414-L415). See for example https://github.com/openjdk/lilliput/blob/daea5edef6eaf56bef1ac157fa872d7f81896c27/src/hotspot/share/gc/shared/gcInitLogger.cpp#L93. And maybe combine with the existing `Number of regions:` log at the end of the function?

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
```diff
@@ -420,6 +422,22 @@ void PSParallelCompactNew::setup_regions_parallel() {
...
+  size_t total_heap_words = 0;
+  for (uint i = old_space_id; i < last_space_id; ++i) {
+    total_heap_words += _space_info[i].space()->capacity_in_words();
+  }
+
+  // Per-worker region count for dynamic region sizing
+  const uint RegionsPerWorker = 20;  // Based on 5% boundary waste threshold
+  const uint max_workers = ParallelScavengeHeap::heap()->workers().max_workers();
+  const size_t total_regions_count = (size_t)max_workers * RegionsPerWorker;
+  const size_t DYNAMIC_REGION_SIZE_WORDS = round_up_power_of_2(total_heap_words / total_regions_count);
```
olivergillespie: Theoretically total_heap_words / total_regions_count could be zero, which round_up_power_of_2 does not support. Perhaps `round_up_power_of_2(MAX2(total_heap_words / total_regions_count, floor_region_size_words))`

rudraroop: Yup, makes sense. Moved the round_up_power_2 usage to the end to avoid hitting the > 0 assert. 

Although I wonder if we're just being defensive here - getting `total_heap_words == 0` would mean you may have bigger problems than the region size calculation

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
rudraroop: Thanks, that's much cleaner. Combined with the existing log at the end

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
rudraroop: Let me try and write a reproducer which can give us visibility into this

rudraroop: Yup, I checked `capacity_in_words` does change as heap shrinks/grows, so it's better to recalculate. I guess it's ok for the definition of humongous to change with time too, since it will be in relation to the heap overall. It's not perfect, but not counter-intuitive either

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
olivergillespie: I don't mean total_heap_words could be 0, but the result of `total_heap_words / total_regions_count` could be 0. Still a very weird situation, but slightly more plausible. Minimum initial heap size at the moment seems to be 1MB = 66k words. So if you had 4,000 workers (!), 4,000 x 20 regions per worker = 80k regions. 66k/80k = 0. Still a bit absurd but worth covering.

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
```diff
@@ -374,7 +374,8 @@ void PSParallelCompactNew::post_compact()
 }
 
 void PSParallelCompactNew::setup_regions_parallel() {
-  static const size_t REGION_SIZE_WORDS = (SpaceAlignment / HeapWordSize);
+  const size_t REGION_SIZE_WORDS = compute_region_size();
```
olivergillespie: not static now, so not all caps?

olivergillespie:
One last style comment but otherwise looks good to me. @rkennke would you be able to take a look at some point? We've seen some real applications suffer from the inability to compact humongous objects, I suspect this would help a lot in practice.

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
rudraroop: So the style guide says 
```
Constant names may be upper-case or mixed-case, according to historical necessity.
```
I chose to keep REGION_SIZE_WORDS capital because it's still a const and I guess "historically" it has been all caps. Does that make sense?

olivergillespie: I don't think it counts as a constant now, it's just a local variable like any other after the change. E.g. below:

```
MutableSpace* const space = _space_info[i].space();
```

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
```diff
@@ -417,7 +418,23 @@ void PSParallelCompactNew::setup_regions_parallel() {
     }
   }
   _num_regions = region_idx;
-  log_info(gc)("Number of regions: %zu", _num_regions);
+  log_info(gc)("Number of regions: %zu, region size: " EXACTFMT, _num_regions, EXACTFMTARGS(REGION_SIZE_WORDS * HeapWordSize));
+}
+
+size_t PSParallelCompactNew::compute_region_size() {
```
olivergillespie: Just noticed while reading the style guide it says to avoid `compute`. So, just `region_size()`?

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
```diff
@@ -417,7 +418,23 @@ void PSParallelCompactNew::setup_regions_parallel() {
...
+  // Minimum 0.5MB region size
+  const size_t floor_region_size_words = (SpaceAlignment / HeapWordSize);
```
olivergillespie: This and regions_per_worker could be static.

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
rkennke: This is news to me. I have no problem with `compute_region_size()`. Only `region_size()` I would assume that it is a trivial accessor, which it is not. Is there a reason why 'compute' is discouraged? would `calculate_region_size()` be better?

rkennke (approved):
Changes look good to me, and that is a nice improvement! Thank you!

src/hotspot/share/gc/parallel/psParallelCompactNew.cpp
olivergillespie: > When choosing names, avoid categorical nouns like "variable", "field", "parameter", "value", and verbs like "compute", "get". (storeValue(int param) is bad.)

Not super clear *why* compute is not liked. I agree that keeping it gives a better sense of being a non-trivial calculation so I'm also fine with it, I was just sticking closely to the guide as a precaution. There are [plenty](https://github.com/search?q=repo%3Aopenjdk%2Fjdk%20compute_&type=code) of existing usages, anyway.

olivergillespie (approved):

rkennke (approved):

olivergillespie (approved):

PR #214, opened 2026-04-23
8382631: [Lilliput2] Missing GC barrier in Object::clone

olivergillespie:
The fix for 8372151 split BarrierSetC2::clone into two parts, when using CompactObjectHeaders and cloning an instance.

1. Copy the first 4 bytes of the object, which are not 8-byte aligned (bytes 0-4 are the object header, bytes 4-8 are the first 4 bytes of the object body).
2. Copy the rest of the object with 8-byte copy routine.

This was to avoid copying bytes 0-4 (the header) to the cloned object. In Lilliput 2, this header is stateful for each instance and should not be copied to the clone. The problem is that this 4-byte prefix copy happens outside GC barriers.

The fix here moves the 4-byte copy from clone to clone_at_expansion, where it can be appropriately covered by barriers.

G1: uses `clone_in_runtime` or `BarrierSetC2::clone_at_expansion`
Parallel/Serial: no barriers
Shenandoah: the 4-byte copy is now done after the GC barrier which fixes up all oops on the object.
ZGC: uses `clone_in_runtime` for all instances (arrays are not affected by this change)

We modify the usage of `ArrayCopyNode` slightly - users cannot rely on the offset being long-aligned, which it previously was (for these specific users who do bulk long copies). Now, if the offset is not aligned, the user must handle the int prefix (for which we introduce a helper - `arraycopy_copy_int_prefix`). The `length` of the `ArrayCopyNode` in these cases remains as the number of long elements to copy, i.e. excluding the prefix.

`clone_in_runtime` never suffered from the original problem (copying stateful header), because it explicitly clears the header *after* the copy. But it did rely on the length being rounded *up* to the next long. So, a small adjustment there is needed to round up if offset is not aligned.

Reviewers: please note that this is my first C2 code and it's not based on any theory or deep understanding, I mostly copied from nearby similar patterns (especially `PhaseMacroExpand::generate_block_arraycopy()` which does a similar prefix copy). In particular I don't know if my use of the `MergeMemNode` is correct.

## Testing

Passes the reproducer which crashed before (Dacapo xalan).

---------

---------
### Progress

### Issue
 * [JDK-8382631](https://bugs.openjdk.org/browse/JDK-8382631): [Lilliput2] Missing GC barrier in Object::clone (**Bug** - P4)

### Reviewers
 * [Roman Kennke](https://openjdk.org/census#rkennke) (@rkennke - **Reviewer**) ⚠️ Review applies to [3ca23557](https://git.openjdk.org/lilliput/pull/214/files/3ca235572eeeadddd1d7a299fcfb34cd37ab209f)
 * [Roland Westrelin](https://openjdk.org/census#roland) (@rwestrel - Committer) ⚠️ Review applies to [942f606b](https://git.openjdk.org/lilliput/pull/214/files/942f606bf98bbfc8c159877f967cf945a327a43f)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/214/head:pull/214` \
`$ git checkout pull/214`

`$ git checkout pull/214` \
`$ git pull https://git.openjdk.org/lilliput.git pull/214/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 214`

`$ git pr show -t 214`

</details>

<a href="https://git.openjdk.org/lilliput/pull/214.diff">https://git.openjdk.org/lilliput/pull/214.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/214#issuecomment-4305509403)
</details>

### DIFF
--- a/src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
+++ b/src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2025, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
+++ b/src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
@@ -913,52 +913,52 @@ void BarrierSetC2::clone_at_expansion(PhaseMacroExpand* phase, ArrayCopyNode* ac
 }
 
 bool BarrierSetC2::should_copy_int_prefix(PhaseMacroExpand* phase, ArrayCopyNode* ac) const {
-    // We do our bulk copy in longs. If base offset is not aligned, then we must copy the prefix separately.
-    // With CompactObjectHeaders, the base offset for an instance is 4 bytes.
-    // We cannot simply expand the copy to the previous long-alignment, as that will copy the object header,
-    // which is stateful with COH - it contains hash and lock bits that are specific to the instance.
+  // We do our bulk copy in longs. If base offset is not aligned, then we must copy the prefix separately.
+  // With CompactObjectHeaders, the base offset for an instance is 4 bytes.
+  // We cannot simply expand the copy to the previous long-alignment, as that will copy the object header,
+  // which is stateful with COH - it contains hash and lock bits that are specific to the instance.
 
-    // Skip this when src has an array type. With StressReflectiveCode, the
-    // instance path of the clone can be live in the IR even when the type system
-    // knows src_base is an array. The pre-copy is unnecessary on such paths (they
-    // are unreachable at runtime), and creating a LoadNode at the array length
-    // offset would assert (LoadRangeNode required).
-    Node* src = ac->in(ArrayCopyNode::Src);
-    if (phase->igvn().type(src)->isa_aryptr()) {
-      return false;
-    }
+  // Skip this when src has an array type. With StressReflectiveCode, the
+  // instance path of the clone can be live in the IR even when the type system
+  // knows src_base is an array. The pre-copy is unnecessary on such paths (they
+  // are unreachable at runtime), and creating a LoadNode at the array length
+  // offset would assert (LoadRangeNode required).
+  Node* src = ac->in(ArrayCopyNode::Src);
+  if (phase->igvn().type(src)->isa_aryptr()) {
+    return false;
+  }
 
-    int base_off = arraycopy_payload_base_offset(ac->is_clone_array());
-    if (is_aligned(base_off, BytesPerLong)) {
-      // We're aligned, no need to copy anything separately.
-      return false;
-    }
+  int base_off = arraycopy_payload_base_offset(ac->is_clone_array());
+  if (is_aligned(base_off, BytesPerLong)) {
+    // We're aligned, no need to copy anything separately.
+    return false;
+  }
 
-    guarantee(UseCompactObjectHeaders, "non-aligned base offset only possible with compact object headers");
-    guarantee(is_aligned(base_off, BytesPerInt), "must be 4-bytes aligned");
-    return true;
+  guarantee(UseCompactObjectHeaders, "non-aligned base offset only possible with compact object headers");
+  guarantee(is_aligned(base_off, BytesPerInt), "must be 4-bytes aligned");
+  return true;
 }
 
 MergeMemNode* BarrierSetC2::arraycopy_copy_int_prefix(PhaseMacroExpand* phase, Node* ctrl, Node* mem, Node* src, Node* dst) const {
-    // Manual load/store of one int.
-    MergeMemNode* mm = phase->transform_later(MergeMemNode::make(mem))->as_MergeMem();
-    const TypePtr* s_adr_type = phase->igvn().type(src)->is_ptr();
-    const TypePtr* d_adr_type = phase->igvn().type(dst)->is_ptr();
-    uint s_alias_idx = phase->C->get_alias_index(s_adr_type);
-    uint d_alias_idx = phase->C->get_alias_index(d_adr_type);
-    // This copies the first 4 bytes after the compact header (hash field or first instance field) as a raw int.
-    // The actual field at this offset may be a narrowOop, so the load/store must be marked as mismatched to
-    // avoid StoreN-vs-StoreI assertion failures during IGVN.
-    Node* load_prefix = phase->transform_later(
-        LoadNode::make(phase->igvn(), ctrl, mm->memory_at(s_alias_idx), src, s_adr_type,
-                        TypeInt::INT, T_INT, MemNode::unordered, LoadNode::DependsOnlyOnTest,
-                        false /*require_atomic_access*/, false /*unaligned*/, true /*mismatched*/));
-    Node* store_prefix = phase->transform_later(
-        StoreNode::make(phase->igvn(), ctrl, mm->memory_at(d_alias_idx), dst, d_adr_type,
-                        load_prefix, T_INT, MemNode::unordered));
-    store_prefix->as_Store()->set_mismatched_access();
-    mm->set_memory_at(d_alias_idx, store_prefix);
-    return mm;
+  // Manual load/store of one int.
+  MergeMemNode* mm = phase->transform_later(MergeMemNode::make(mem))->as_MergeMem();
+  const TypePtr* s_adr_type = phase->igvn().type(src)->is_ptr();
+  const TypePtr* d_adr_type = phase->igvn().type(dst)->is_ptr();
+  uint s_alias_idx = phase->C->get_alias_index(s_adr_type);
+  uint d_alias_idx = phase->C->get_alias_index(d_adr_type);
+  // This copies the first 4 bytes after the compact header (hash field or first instance field) as a raw int.
+  // The actual field at this offset may be a narrowOop, so the load/store must be marked as mismatched to
+  // avoid StoreN-vs-StoreI assertion failures during IGVN.
+  Node* load_prefix = phase->transform_later(
+      LoadNode::make(phase->igvn(), ctrl, mm->memory_at(s_alias_idx), src, s_adr_type,
+                      TypeInt::INT, T_INT, MemNode::unordered, LoadNode::DependsOnlyOnTest,
+                      false /*require_atomic_access*/, false /*unaligned*/, true /*mismatched*/));
+  Node* store_prefix = phase->transform_later(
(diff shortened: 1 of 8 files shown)


olivergillespie:
~~I'm seeing a crash with fastdebug and ZGC - runtime/cds/appcds/TestZGCWithCDS.java. Looking into it.~~

That fails on `master`, too, so not caused by this change. It looks like a simple issue with `AOTStreamedHeapWriter::update_header_for_buffered_addr`, `initialize_hash_if_necessary` is called *before* the `set_narrow_klass`, but `initialize_hash_if_necessary` has an assertion that `narrowKlass` is set. Probably an easy fix.

rkennke (commented):
Hi @olivergillespie, this is a good change and I *think* it is correct. It would be good to have a C2 expert have a look, too, though. Maybe @rwestrel or @JohnTortugo can have a look if the new wiring is correct?

Also, you should definitely run more tests on it: Check that the new test from https://github.com/openjdk/lilliput/pull/207 is still ok, run tier1 (and ideally also tier2-3) with and without +UCOH. If it's possible to come up with a regression test for the specific scenario that you are fixing, this would be perfect!

olivergillespie:
Thanks. Yes, please, any more eyes very much appreciated.

I'm running those tests now. As for the regression test, I tried for quite a while at the start of the bug - it wasn't too hard to trigger a bad copy, but I never actually triggered a crash, so I kept Dacapo xalan as my true reproducer. But now that I feel I understand it even more, I'll try again for a tight reproducer. The key annoyance was the tight Shenandoah window when the bad pointers are actually in place, are there any existing tests or techniques to stress that that you know of?

olivergillespie:
I've run tiers 1-4 with/without COH, and with Shenandoah, all on x86 only. No *new* failures. The Dacapo reproducer no longer crashes, and TestStressIHash still passes. So it looks good from that perspective.

I'm still trying to create a tight reproducer, it's more difficult than I thought, I think because a lot of the failures are benign - if a from-space pointer is copied to a new object without resolving, it will usually just be resolved on the next load or by update-refs. So I'm looking at a scenario where C2 omits LRB on the cloned object too, because it's 'freshly allocated' - in that scenario we could see problems (but I haven't, yet).

I could maybe add a separate dedicated assertion - after cloning, walk the object and verify no from-space refs. That would be a lot more reliable, obviously.

rkennke:
> I've run tiers 1-4 with/without COH, and with Shenandoah, all on x86 only. No _new_ failures. The Dacapo reproducer no longer crashes, and TestStressIHash still passes. So it looks good from that perspective.
> 
> I'm still trying to create a tight reproducer, it's more difficult than I thought, I think because a lot of the failures are benign - if a from-space pointer is copied to a new object without resolving, it will usually just be resolved on the next load or by update-refs. So I'm looking at a scenario where C2 omits LRB on the cloned object too, because it's 'freshly allocated' - in that scenario we could see problems (but I haven't, yet).
> 
> I could maybe add a separate dedicated assertion - after cloning, walk the object and verify no from-space refs. That would be a lot more reliable, obviously.

Have you tried -XX:+ShenandoahVerify? I think it would be hard to verify purely from Java, for the reasons that you mention. But if you *avoid* accessing the cloned arrays (thus avoiding to update the refs by the LRB), then +ShenandoahVerify might catch the from-space refs. Also, -XX:ShenandoahGCHeuristics=aggressive might help.

olivergillespie:
I have tried ShenandoahVerify and have set up the test so that we're definitely cloning while the barrier is active. But I don't think ShenandoahVerify actually covers our case. ~~You can even set `-XX:-ShenandoahCloneBarrier` without any obvious failures.~~

The actual need for a clone barrier seems very narrow. I can see two options:
1. The compiler relies on the to-space invariant, so elides future barriers causing real problems.
2. A possible scenario where update-refs does not walk the new object to fix its refs, because it's above the watermark (still trying to understand if this is possible).
I'm reading https://bugs.openjdk.org/browse/JDK-8231086 and follow-ups to try to understand exactly why clone barrier is important. 

Edit: actually -Barrier +Verify does fail on TestStressIHash.java, so I'll chase that.

olivergillespie:
Okay I seem to have a reproducer. I'll tidy it up and add it.

olivergillespie:
The reproducer is:
```java
import java.util.*;

static class Instance implements Cloneable {
    Object ref = new Object();

    public Instance clone() throws CloneNotSupportedException {
        return (Instance) super.clone();
    }
}

void main(String[] args) throws Exception {
    List<Instance> instances = new ArrayList<>();

    for (int i = 0; i < 100_000; i++) {
        instances.add(new Instance());
        Instance instance = instances.get(i % 1000);
        Instance clone = (Instance) instance.clone();
    }    
}
```

```
java -XX:+UseShenandoahGC -Xmx100M -XX:ArrayCopyLoadStoreMaxElem=0 \
-XX:+UseCompactObjectHeaders -XX:+UnlockDiagnosticVMOptions \
-XX:+ShenandoahVerify -XX:ShenandoahGCHeuristics=aggressive CloneRepro.java
```

Fails pretty reliably without the fix, and never with it. I'm trying to add it to `test/hotspot/jtreg/gc/shenandoah/compiler/TestClone.java` but it's less reliable there for some reason, still figuring that out.

olivergillespie:
I am adding the reproducer to TestClone.java upstream (https://github.com/openjdk/jdk/pull/31001), it's reliable now (just needed aggressive+verify), so I think we can skip it here and wait for the merge. It reliably fails without the fix, and passes with it.

rkennke:
> I am adding the reproducer to TestClone.java upstream ([openjdk/jdk#31001](https://github.com/openjdk/jdk/pull/31001)), it's reliable now (just needed aggressive+verify), so I think we can skip it here and wait for the merge. It reliably fails without the fix, and passes with it.

That sounds fine.
Just make sure we have C2 folks look at it.

src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
```diff
@@ -897,11 +858,11 @@ void BarrierSetC2::clone_in_runtime(PhaseMacroExpand* phase, ArrayCopyNode* ac,
 
   // The native clone we are calling here expects the object size in words.
   // Add header/offset size to payload size to get object size.
-  // Use the actual offset stored in the ArrayCopyNode (in bytes), not
-  // arraycopy_payload_base_offset(), because clone() may have bumped the
-  // offset past a 4-byte pre-copy for compact object headers.
-  Node* const base_offset = phase->transform_later(new URShiftXNode(ac->in(ArrayCopyNode::SrcPos), phase->intcon(LogBytesPerLong)));
-  Node* const full_size = phase->transform_later(new AddXNode(size, base_offset));
+
+  // If not aligned, round *up*.
```
rwestrel: Is there not risk to read/write beyond the end of the heap if that object is right at the end of the heap?

src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
```diff
@@ -942,6 +912,55 @@ void BarrierSetC2::clone_at_expansion(PhaseMacroExpand* phase, ArrayCopyNode* ac
...
+      return false;
+    }
+
+    int base_off = arraycopy_payload_base_offset(ac->is_clone_array());
+    if (is_aligned(base_off, BytesPerLong)) {
+      // We're aligned, no need to copy anything separately.
+      return false;
+    }
+
+    guarantee(UseCompactObjectHeaders, "non-aligned base offset only possible with compact object headers");
```
rwestrel: assert here and below would be good enough, I think.

rwestrel (approved):
The c2 side of things looks reasonable to me.

src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
olivergillespie: No, this never reads past object end. ArrayCopyNode length is payload size in longs, not including header, and also not including the possible unaligned first 4 bytes of the payload.

Runtime clone copies from 0, not from offset, so the length is too short, we need to add the base_offset, plus the 4-byte prefix if applicable.

For an aligned offset, there's no difference from before, rounding up doesn't change anything.

```
[0       ][4        ][8        ][12       ]
[markword][arylength][ary elems (+padding)]

length = 1 long, offset = 8 bytes = 1 long
full_size = 1 + 1 = 2
copy(offset=0, size=2 longs)
```

For an unaligned offset, we must round up.

```
[0       ][4        ][8        ][12       ]
[markword][object body          (+padding)]

length = 1 long, offset = 4 bytes = 1 long (must round *up* to 1 long here to get the right total size)
full_size = 1 + 1 = 2
copy(offset=0, size=2 longs)
```

This always gets us to the full aligned size of the object, no more. The resulting length is the same as before, it's just that the rounding up has moved places. Before, it was rounded up by bumping base_offset in `clone` after copying the prefix, so clone_in_runtime always saw an aligned offset.

```
// Before PR 214

[0       ][4        ][8        ][12       ]
[markword][object body          (+padding)]

clone:
offset =  4 bytes
copy(offset=4 bytes, size = 4bytes)
offset += 4 bytes
new ArrayCopyNode(offset=8 bytes, length=1 long)
...
clone_in_runtime:
offset = 8 bytes (= 1 long)
full_size = length + 1
copy(offset=0, 2 longs)

```

src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
olivergillespie: Thanks. The previous code had guarantees so I just continued like that. They look cheap enough that I'm not worried about the overhead, but I don't have a strong preference. @rkennke do you have an opinion?

olivergillespie:
> The c2 side of things looks reasonable to me.

Thanks a lot for reviewing!

src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
olivergillespie: That was a long way of saying that the logic is:

```
clone_in_runtime:
   full_size = align(offset) + (payload_size)
   copy(offset=0, full_size)
   init_header()
```

So it always copies the full size of the object, it's just details on how to re-calculate that from the ArrayCopyNode.

olivergillespie: Added a small comment that is hopefully clearer.

olivergillespie:
Made some minor tweaks (copyright, indenting, added const, removed virtual, updated two comments). No functional change. Sorry for the churn, missed these earlier.

src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
rkennke: No I don't mind either way.

rkennke (changes requested):
Looks good to me, thank you!

rkennke (approved):
Oops clicked wrong button.

src/hotspot/share/gc/shared/c2/barrierSetC2.cpp
olivergillespie: okay, switched to asserts then per Roland's preference

olivergillespie:
Great, thanks both!

/integrate

PR #221, opened 2026-06-26
8387335: [Lilliput2] G1: Preserve hash after humongous compaction

olivergillespie:
Fix for G1 humongous object compaction with Lilliput 2 related to hash-code expansion.

G1 correctly uses copy_size in the compaction planning phase, but then when it comes to actually moving it, we don't account for copy_size. In particular, passing a too-small size to set_humongous_metadata can cause the filler object to overwrite the hash-code at the end of the object.

Also fix the region accounting (note that skipping the fix for `free_non_overlapping_regions` doesn't cause the test to fail, I'm not sure exactly why. it fails when removing any of the other fixes, though).

Before the fix, the test will fail with:

```
# JRE version: OpenJDK Runtime Environment (27.0) (build 27-internal-adhoc.ogillesp.lilliput)
# Java VM: OpenJDK 64-Bit Server VM (27-internal-adhoc.ogillesp.lilliput, mixed mode, sharing, tiered, compressed oops, compact obj headers, g1 gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0x96ebdd][0.456s][error][gc,verify] GC(25) Object 0x00000000e1080018 has a null klass
```

or if we don't use VerifyDuringGC, then we'll see:

```
...
hash mismatch at 509
hash mismatch at 511
hash mismatch at 517
hash mismatch at 519
STDERR:
java.lang.RuntimeException: Identity hash-code, length or data corruption detected.
```

The test is intentionally set up to be ready for Shenandoah, too, which has a similar issue.

---------

---------
### Progress
- \[x\] Change must not contain extraneous whitespace
- \[x\] Commit message must refer to an issue
- \[x\] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.org/bylaws#committer))

### Issue
 * [JDK-8387335](https://bugs.openjdk.org/browse/JDK-8387335): [Lilliput2] G1: Preserve hash after humongous compaction (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/221/head:pull/221` \
`$ git checkout pull/221`

`$ git checkout pull/221` \
`$ git pull https://git.openjdk.org/lilliput.git pull/221/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 221`

`$ git pr show -t 221`

</details>

<a href="https://git.openjdk.org/lilliput/pull/221.diff">https://git.openjdk.org/lilliput/pull/221.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/221#issuecomment-4810622388)
</details>

### DIFF
--- a/src/hotspot/share/gc/g1/g1Allocator.cpp
+++ b/src/hotspot/share/gc/g1/g1Allocator.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014, 2025, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2014, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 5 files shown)


rkennke (approved):
Looks good to me. Can you please hold off intergrating until the merge-28+2 lands (later today)?

rkennke (approved):
Good catch and changes still look good!
I now integrated merge-jdk-28+2 and you can integrate this PR and also the other that you have lined up.

PR #223, opened 2026-06-30
8387285: [Lilliput2] Shenandoah: Preserve hash after humongous compaction

pengxiaolong:
The PR is based on @olivergillespie's PR [here](https://github.com/olivergillespie/lilliput/pull/1), with some additional work to fix the problem in TLAB allocation when the tlab size is same as region size, along with new test to cover the edge case.

Test:
- [x] hotspot_gc_shenandoah

---------

---------
### Progress
- \[x\] Change must not contain extraneous whitespace
- \[x\] Commit message must refer to an issue
- \[x\] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.org/bylaws#committer))

### Issue
 * [JDK-8387285](https://bugs.openjdk.org/browse/JDK-8387285): [Lilliput2] Shenandoah: Preserve hash after humongous compaction (**Bug** - P4)

### Reviewers
 * [Oli Gillespie](https://openjdk.org/census#ogillespie) (@olivergillespie - Committer) ⚠️ Review applies to [1b7e61a2](https://git.openjdk.org/lilliput/pull/223/files/1b7e61a29e9879af3d431a7c37cabf0132a25d58)
 * [Roman Kennke](https://openjdk.org/census#rkennke) (@rkennke - **Reviewer**) ⚠️ Review applies to [857599b5](https://git.openjdk.org/lilliput/pull/223/files/857599b50ac7e4c6d5b93fce6227869762e30c3a)

### Contributors
 * Oli Gillespie `<ogillespie@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/lilliput.git pull/223/head:pull/223` \
`$ git checkout pull/223`

`$ git checkout pull/223` \
`$ git pull https://git.openjdk.org/lilliput.git pull/223/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 223`

`$ git pr show -t 223`

</details>

<a href="https://git.openjdk.org/lilliput/pull/223.diff">https://git.openjdk.org/lilliput/pull/223.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/lilliput/pull/223#issuecomment-4845286075)
</details>


src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.hpp
```diff
@@ -301,7 +301,23 @@ class ShenandoahHeapRegion {
...
+  // freshly allocated, not-yet-hashed object that could later grow by one word
+  // when an identity hash-code is injected as it is copied during GC (see
+  // oopDesc::copy_size). In that case we reserve the extra word up front so the
+  // object is routed to the humongous path if its expanded form would overflow a
+  // region. It must be false (the default) for every size that is already final:
+  //   - the live size of an existing object (obj->size()), which already includes
+  //     the hash word once the object has been expanded by a prior GC copy;
+  //   - an evacuation/promotion copy size (copy_size()), which is already expanded;
+  //   - a raw LAB buffer (TLAB/GCLAB/PLAB), which never grows.
+  inline static bool requires_humongous(size_t words, bool may_expand_for_hash = false) {
```
olivergillespie: Would it be worth removing the default so that callers are forced to understand and choose? To prevent mistakes. Seems like there aren't many callers so it won't make a big mess.

test/hotspot/jtreg/gc/shenandoah/options/TestCompactHeaderTLABSize.java
```diff
@@ -0,0 +1,90 @@
...
+    public static void main(String[] args) throws Exception {
+        // With compact object headers, the max TLAB is capped one word below the
+        // region, so requesting a region-sized TLAB must be rejected at startup.
+        OutputAnalyzer coh = ProcessTools.executeLimitedTestJava(
+                "-XX:+UnlockExperimentalVMOptions",
+                "-XX:+UnlockDiagnosticVMOptions",
+                "-XX:+UseShenandoahGC",
+                "-XX:+UseCompactObjectHeaders",
+                "-XX:ShenandoahRegionSize=" + REGION_SIZE,
+                "-XX:TLABSize=" + REGION_TLAB,
```
olivergillespie: Minor, I'd find TLABSize=REGION_SIZE clearer.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -3128,7 +3128,13 @@ void ShenandoahFreeSet::decrease_humongous_waste_for_regular_bypass(ShenandoahHe
 
 HeapWord* ShenandoahFreeSet::allocate(ShenandoahAllocRequest& req, bool& in_new_region) {
   shenandoah_assert_heaplocked();
-  if (ShenandoahHeapRegion::requires_humongous(req.size())) {
+  // A freshly allocated mutator object is not yet hashed and may grow by one word
+  // when an identity hash-code is injected during a future GC copy, so reserve that
+  // headroom up front. Every other request carries an already-final size: a GC copy
```
olivergillespie: Hmm, are we actually reserving the headroom? Or just going down the humongous path?

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.hpp
pengxiaolong: Make sense, I'll remove the default and enforce callers to provide the argument.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
pengxiaolong: No, we are not, I'll update the comments.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.hpp
pengxiaolong: You are right, the default is error-prone, also fixed few more potential assert from the call sites, they may cause assert if we design specific test.

olivergillespie (approved):

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.hpp
```diff
@@ -301,7 +301,23 @@ class ShenandoahHeapRegion {
...
+  // oopDesc::copy_size). In that case we reserve the extra word up front so the
+  // object is routed to the humongous path if its expanded form would overflow a
+  // region. It must be false for every size that is already final:
+  //   - the live size of an existing object (obj->size()), which already includes
+  //     the hash word once the object has been expanded by a prior GC copy;
+  //   - an evacuation/promotion copy size (copy_size()), which is already expanded;
+  //   - a raw LAB buffer (TLAB/GCLAB/PLAB), which never grows.
+  inline static bool requires_humongous(size_t words, bool may_expand_for_hash) {
+    if (UseCompactObjectHeaders && may_expand_for_hash) {
+      words = align_object_size(words + 1);
```
rkennke: Can you add a comment noting to keep that code in sync with `markWord::copy_size()`? Or avoid that implicit coupling altogether by making it explicit by introducing a tiny helper method, e.g.:

```
static size_t hash_expanded_size(size_t words) { return align_object_size(words + 1); }
```

and make both places use that.

rkennke (approved):
Looks good to me, just a tiny nit. Not worth another review.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.hpp
pengxiaolong: Thanks @rkennke! 
I have added the helper method, and replaced all the places to use the helper method(there are 4 more places).

